Issue #4, opened 2015-06-06
react-template-helper doesn't un-mount React component


AdamBrodzinski:
Does Meteor automatically clean up the React component that get's rendered in with the `{{React } }` helper? It seems like the current build will leak once it's torn down.

I was going to try and submit a PR but I couldn't figure out a way to reference the node once `onDestroyed` is called since `this.firstNode` is null.

I'm manually doing this with an ID like so and that works:

```
Template.postsNew.rendered = function() {
  var container = document.getElementById('PostContainer');
  React.render(React.createElement(PostForm), container);
};

Template.postsNew.destroyed = function() {
  var container = document.getElementById('PostContainer');
  React.unmountComponentAtNode(container);
};
```

avital:
Hmm. The id approach won't work if you reuse the component. You're right that we need to implement this. It's unfortunate that there's no `beforeDestroyed` hook but maybe there's a way to get around this by accessing the view directly... I'll put this on my list, thanks.

avital:
(See https://github.com/meteor/meteor/issues/2141 for the `beforeDestroyed` feature request)

avital:
Thanks for the pull request, but I found a simpler solution that also doesn't add unnecessary attributes to the DOM.

PR #8, opened 2015-06-09
Fixes Issue #4 -  Leaking React Components

AdamBrodzinski:
Patch for issue #4 where a React component doesn’t get un-mounted
before  Blaze destroys it’s own view.

Creates a random id and attaches it to the blaze wrapper container with
a data attribute. This value is stored in the template instance so it
can be accessed at `onDestroy`.

Sample repo can be used to verify proper use. Inspector shows an empty
Blaze container div after un-mounting and before exiting `onDestroyed`

https://github.com/AdamBrodzinski/react-blaze-helper-fix

`PACKAGE_DIRS=` was used to link this patch to repo above.

### DIFF
--- a/react-template-helper/react-template-helper.js
+++ b/react-template-helper/react-template-helper.js
@@ -1,6 +1,8 @@
 Template.React.onRendered(function () {
   var parentTemplate = parentTemplateName();
   var container = this.firstNode.parentNode;
+  this._blazeId = Math.random().toString();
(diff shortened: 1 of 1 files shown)


avital:
Fixed with a simpler solution, and added tests in https://github.com/meteor/react-packages/commit/e964e14478e2ddc79944b814af7db0bdf7f8e962

PR #18, opened 2015-06-25
Create a version of Meteor platform without Blaze

stubailo:
Also, implement a static HTML package that just lets you define static content for `<head>` and `<body>`.

Add both of these to Todos.


examples/react-todos/client/routes.jsx
```diff
@@ -39,7 +39,8 @@ Meteor.startup(function () {
       showFirstList();
     }
 
-    React.render(<Handler handles={ handles } />, document.body);
+    React.render(<Handler handles={ handles } />,
+      document.getElementById("app-container"));
```
avital: Just skimming, but why can't we render to `document.body` in this case?

examples/react-todos/.meteor/packages
```diff
@@ -13,5 +14,6 @@ reactive-var
 meteorhacks:npm
 cosmos:browserify
 
+static-html
```
avital: Should this be part of `meteor-platform-no-ui`?

examples/react-todos/.meteor/versions
```diff
@@ -60,6 +59,7 @@ sha@1.0.3
 spacebars@1.0.6
 spacebars-compiler@1.0.6
 srp@1.0.3
+static-html@0.0.1
 templating@1.1.1
```
avital: Hmm, why is templating still here?

avital:
Just skimmed, but looks promising!

Are you duplicating code from the templating package? Can we instead somehow share code?

examples/react-todos/client/routes.jsx
AdamBrodzinski: @avital one of the subtle issues with rendering directly into `document.body` is that 3rd party scripts (like analytics, social buttons, etc...) can inject nodes into `document.body` and React doesn't seem to like it. This can cause hard to track down bugs.

examples/react-todos/.meteor/packages
stubailo: It depends on whether we want this to be React-specific, or also work for Angular. I was thinking Angular could then finally use the `.html` file extension. @Urigo, what do you think? If there was a `meteor-platform` without the HTML source handler, would that make the Angular experience better?

examples/react-todos/.meteor/versions
stubailo: It's definitely not running inside the app, because then `static-html` wouldn't work since you can't have two handlers for the same extension.

examples/react-todos/.meteor/packages
Urigo: @stubailo  the only reason we use .ng.html is to get something like that, so I guess our HTML source handler will probably be exactly like `static-html`

Urigo: from looking at the code it doesn't seem like `static-html` is React specific, or I'm missing something?

Urigo: another thing would be to create a client only package that we could serve from a static website and not from Meteor, but I guess that is out of scope (although might be important for React devs as well)

stubailo: Oh, does the .ng.html thing not do anything special with the Angular templates? I thought maybe you would want to have that act on the .html files directly.

stubailo: If static-html will work for React and Angular both, then it should probably be in the meteor platform

Urigo: Yes I think it should be inside.
Right now we use `ng.html` to minify and also add the template to `template-cache` but it will be better that it will be optional.
Developers could use regular HTML files and if they really want they would use `ng.html` for template cache. Right now everything is in template cache.

By the way, what about minification and other plugins?

https://github.com/Urigo/angular-meteor/blob/master/plugin/handler.js

stubailo: Since this is static HTML only, it doesn't have anything to minify? Not sure how it is related.

Urigo: I mean like https://github.com/kangax/html-minifier
but I guess we Meteor doesn't do it anyway so never mind...

stubailo: Well, there is always room for improvement! We could minify the HTML if necessary.

Urigo: It might be a good idea, what do you think?

Urigo: It's definitely not necessary, nice to have at most

stubailo: Yeah, I think that would be a separate project.

stubailo:
@avital after talking to @Slava, I have discovered that the `templating` package might be going through a lot of changes in the next Meteor release. Since I would like to ship the `static-html` package in the next week (possibly even on Monday) I think it would be best to do a hacky code-duplicating version in the mean time, and then refactor when the batch plugin changes land.

The only other alternative I see is delaying shipping this package until batch plugins land.

stubailo:
Rebased on newest stuff. I'm still not sure if I should add `static-html` to the `meteor-platform-no-ui`... would people want to add their own source handlers for `html`?

avital:
I vote for `meteor-platform-no-ui` to /not/ include `static-html`. Since otherwise people will similarly complain that they can't remove our default html parsing with their own custom one of they want to.

stubailo:
After talking with @avital, I've decided that it would be best to ship these packages as part of Meteor 1.2 (and not before) in order to take advantage of the ongoing work on build system improvements.

Issue #23, opened 2015-06-27
Tutorial places app in document.body


AdamBrodzinski:
Putting the root React app into `document.body` can cause a lot of unforeseen problems with browser extensions and other JS that append to body. [The tutorial](https://github.com/meteor/react-packages/blob/master/docs/tutorial.md#step-2-adding-your-first-component) shows you how to render to body. Should we change this or at least provide a warning to not do this in production?

Perhaps something like:

```
<body>
  <div id="root"></div>
</body>
```
and
```
if (Meteor.isClient) {
  Meteor.startup(function () {
    React.render(<App />, document.getElementById('root'));
  });
}     
```

stubailo:
Good point - if you submit a PR I'll accept it, or I'll get to it a bit later.

AdamBrodzinski:
Sure thing, i'll submit one tonight.

PR #24, opened 2015-06-27
Fixes Issue #23 by Mounting to Container Div

AdamBrodzinski:
This commit fixes issue #23 by changing the tutorial to mount to a container div instead.

Putting the root React app into document.body can cause a lot of
unforeseen problems with browser extensions and other JS that append to
body.


stubailo:
Awesome, thanks.

PR #34, opened 2015-06-30
Use meteor-babel NPM package instead of using babel-core directly.

benjamn:
Tested with
```sh
cd path/to/react-packages/packages/babel-tests
PACKAGE_DIRS=.. meteor test-packages ./
```

### DIFF
--- a/packages/babel-compiler/babel.js
+++ b/packages/babel-compiler/babel.js
@@ -1,52 +1,30 @@
-Babel = Npm.require('babel-core');
+var meteorBabel = Npm.require('meteor-babel');
 
 // See README.md in this directory for more information.
 
-Babel.transformMeteor = function (code, extraOptions) {
-  extraOptions = _.extend({}, extraOptions); // clone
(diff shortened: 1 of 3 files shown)


packages/babel-compiler/babel.js
```diff
@@ -1,52 +1,30 @@
...
 // See README.md in this directory for more information.
 
-Babel.transformMeteor = function (code, extraOptions) {
-  extraOptions = _.extend({}, extraOptions); // clone
+Babel = {
+  transformMeteor: function (code, extraOptions) {
+    // See https://github.com/meteor/babel/blob/master/options.js if you
+    // want to know what these default options are.
+    var options = meteorBabel.getDefaultOptions();
+    delete options.sourceMap;
```
stubailo: What does this do?

benjamn: The default options have `sourceMap: "inline"`, which breaks the tests in this repository.

PR #46, opened 2015-07-08
Refactor Leaderboard App

AdamBrodzinski:
Hello,

I noticed that some of the examples were not following common React patterns. I refactored the Leaderboard app a bit to conform more closely to React idioms. This changed the structure a little more that I anticipated so I hope it's not going too far. I can also split `PlayerItem` out into another file too.

Would you be open to converting the increment button section to a react template too? This would leave Blaze to handle just the HTML static shell.

Ideally from a React perspective there would be a topmost `LeaderBoardContainer` React component wrapping both the increment component and `PlayerList` component and that would get inserted into the body but I wasn't sure how much of the Blaze logic you wanted to leave.

Cheers!
Adam

### DIFF
--- a/examples/leaderboard/leaderboard.html
+++ b/examples/leaderboard/leaderboard.html
@@ -15,7 +15,7 @@
 
 <template name="leaderboard">
   <div>
-    {{> React component=Leaderboard
+    {{> React component=PlayerList
       players=players.fetch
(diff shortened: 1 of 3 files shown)


stubailo:
Looks great! Some of the spacing between brackets is a by inconsistent, but overall a big improvement.

PR #67, opened 2015-07-20
Update tutorial.md

brennanreese:
Changed the wording on the last line to indicate the link goes to the tutorial and Next will continue taking the user through the preview.


brennanreese:
Changed the wording a bit. If it fooled me, perhaps it will fool others. Or maybe, I'm just a fool!

PR #75, opened 2015-07-29
Bug adding new task in React-ToDos example

jeffmitchel:
The input is not the target event, so new task text is not added properly. Added a ref to the input and use it to retrieve the text.


stubailo:
Oh, good point! I don't know how we missed this.

Can you sign the contributor agreement here? https://contribute.meteor.com/

jeffmitchel:
Signed. Thanks!

PR #77, opened 2015-07-30
Fix react-meteor-data for server rendering

eXon:
When using the mixin `ReactMeteorData`, it is not taking into account that it might be executed on the server. Because it uses the Tracker (and it doesn't work on the server), it throws this error:

```
I20150730-19:37:02.216(-4)? Exception in callback of async function: Error: Can't call yield in a noYieldsAllowed block!
I20150730-19:37:02.217(-4)?     at Function.Fiber.yield (packages/meteor/fiber_helpers.js:8:1)
I20150730-19:37:02.218(-4)?     at Function.wait (/home/benoit/.meteor/packages/meteor-tool/.1.1.3.4sddkj++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:183:14)
I20150730-19:37:02.218(-4)?     at Object.Future.wait (/home/benoit/.meteor/packages/meteor-tool/.1.1.3.4sddkj++os.linux.x86_64+web.browser+web.cordova/mt-os.linux.x86_64/dev_bundle/server-lib/node_modules/fibers/future.js:397:10)
I20150730-19:37:02.218(-4)?     at [object Object]._.extend._nextObject (packages/mongo/mongo_driver.js:929:1)
I20150730-19:37:02.219(-4)?     at [object Object]._.extend.forEach (packages/mongo/mongo_driver.js:963:1)
I20150730-19:37:02.219(-4)?     at [object Object]._.extend.map (packages/mongo/mongo_driver.js:973:1)
I20150730-19:37:02.219(-4)?     at [object Object]._.extend.fetch (packages/mongo/mongo_driver.js:997:1)
I20150730-19:37:02.220(-4)?     at [object Object].Cursor.(anonymous function) [as fetch] (packages/mongo/mongo_driver.js:812:1)
I20150730-19:37:02.221(-4)?     at [object Object].React.createClass.getMeteorData (./app.jsx:30:31)
I20150730-19:37:02.221(-4)?     at packages/react-meteor-data/meteor-data-mixin.jsx:73:28
```

When we are using ReactMeteorData on the server, we should get the `getMeteorData()` straight away. This PR is fixing the problem. Here is a very simple example of server-side rendering with react-router:

```javascript
TestItems = new Mongo.Collection('TestItems');

if (Meteor.isServer) {
  if (TestItems.find().count() === 0) {
    TestItems.insert({ content: 'line 1' });
    TestItems.insert({ content: 'line 2' });
    TestItems.insert({ content: 'line 3' });
  }
}

const {Router, Route, Link} = ReactRouter;

Test = React.createClass({
  render() {
    return (
      <ul>
        {this.props.items.map(function(item) {
          return <li key={item._id}>{item.content}</li>;
        })}
      </ul>
    );
  }
});

App = React.createClass({
  mixins: [ReactMeteorData],

  getMeteorData() {
    return {
      items: TestItems.find().fetch()
    };
  },

  render() {
    console.log(this.data.items);
    return (
      <div>
        <h1>Test!</h1>
        <Test items={this.data.items} />
      </div>
    );
  }
});

AppRoutes = (
  <Route path="/" component={App} />
);

if (Meteor.isClient) {
  const {history} = ReactRouter.lib.BrowserHistory;

  Meteor.startup(function() {
    React.render((
      <Router history={history} children={AppRoutes} />
    ), document.getElementById('react-root'));
  });
}

if (Meteor.isServer) {
  FastRender.route('/', function() {
    this.subscribe('TestItems');
  });

  Meteor.publish('TestItems', function () {
    return TestItems.find();
  });


  const Location = ReactRouter.lib.Location;
  const url = Npm.require('url');

  WebApp.rawConnectHandlers.use(function(req, res, next) {
    var originalWrite = res.write;
    res.write = function(data) {
      if(typeof data === 'string') {
        var parsedUrl = url.parse(req.url);
        var location = new Location(parsedUrl.pathname, parsedUrl.query);

        Router.run(AppRoutes, location, function(error, initialState, transition) {
          var html = '';

          if (initialState) {
            Meteor.bindEnvironment(function() {
              html = React.renderToString(
                <Router {...initialState} children={AppRoutes} />
              );
            })();
          }

          var reactRoot = '<div id="react-root">' + html + '</div>';
          data = data.replace('<body>', '<body>' + reactRoot);
        });
      }

      originalWrite.call(this, data);
    };

    next();
  });
}
```

### DIFF
--- a/packages/react-meteor-data/meteor-data-mixin.jsx
+++ b/packages/react-meteor-data/meteor-data-mixin.jsx
@@ -55,6 +55,12 @@ class MeteorDataManager {
     if (! component.getMeteorData) {
       return null;
     }
+    
+    // When rendering on the server, we don't want to use the Tracker.
+    // We only do the first rendering on the server so we can get the data right away
+    if (Meteor.isServer) {
+      return component.getMeteorData();
+    }
 
     if (this.computation) {
       this.computation.stop();


eXon:
You can now refer to the [react-router-ssr package](https://atmospherejs.com/reactrouter/react-router-ssr) if you would like to test the fix easily.

grigio:
thanks @eXon  +1 in this way Meteor React SSR with data works.

dgreensp:
Looks like you can't "yield" inside a `Tracker.autorun`, and accessing Mongo yields.

Although using Tracker on the server is not very supported at the moment, it feels like it breaks isomorphism to kill `getMeteorData()`'s reactive aspect based on `Meteor.isServer`.  Is there any other way to do it that works for your SSR case?

eXon:
@dgreensp The only purpose of running a React component and ReactMeteorData on the server is for server rendering. I can't see a single use-case where it will be useful (maybe I'm missing something). 

You cannot run a React component without the DOM unless you use renderToString or shallow rendering (this is made strictly for testing).

dgreensp:
Ah I see, I just read up on `renderToString`, which calls `componentWillMount` and then no other callbacks.  So `renderToString` will probably leak autoruns on the client.  Is there any way to know if we are inside `renderToString`?  That aside, though, I'm leaning towards taking this PR now.

eXon:
React is such a different beast than most framework :). In the React community, most people are either leaving what they don't want in the renderToString inside the `componentDidMount()` or checking with `if (typeof window === 'undefined')`. Their is no definitive way to know, but checking `Meteor.isServer` looks like the most sure.

dgreensp:
Ok, sounds good.

PR #88, opened 2015-08-13
Check for Mongo existence outside key checks

rclai:
Quick optimization.

### DIFF
--- a/packages/react-meteor-data/meteor-data-mixin.jsx
+++ b/packages/react-meteor-data/meteor-data-mixin.jsx
@@ -107,16 +107,17 @@ class MeteorDataManager {
         }
       });
     });
-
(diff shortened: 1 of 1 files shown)


stubailo:
Thanks!

Slava:
@stubailo is this change actually correct? I think you are checking against a server-side cursor, not the client-side one (which is LocalCollection.Cursor or something). I think the way `observe-sequence` does it, is just checking the `_publishCursor` method.

stubailo:
Damn.

stubailo:
@Slava wait a second. I'm pretty sure it works because I have tests for it

Slava:
@stubailo OK, great then :)

Issue #116, opened 2015-10-07
Upgrade to React 0.14


stubailo:
http://facebook.github.io/react/blog/2015/10/07/react-v0.14.html

We need to get on top of this! That includes updating examples, etc.

andreasbhansen:
Yes please! :+1:

bwhitty:
I see that the Meteor package versions are tracking React versions, but I'm not sure this makes sense for Meteor. I think the dev/prod runtime packages should probably be bumped to 1.0, and when an update to React 0.14 is pushed, bump them to 2.0 because there are breaking changes introduced in React 0.14. Thoughts?

stubailo:
I think this would be more confusing than helpful - how would you even find out which features of React are available in "Meteor React" version 3.1.4? It could be 0.15, or 1.2, or anything else.

bwhitty:
I guess what I'm really getting at is if the `react` package is updated to React `0.14` and I run a `meteor update`, am I going to unwittingly get BC breaks because this package isn't technically SemVer?

stubailo:
Yep, that's correct. @dgreensp what do you think about this issue?

dgreensp:
Using different major versions would create more problems than it would solve, because of packages that depend on `react`.  Different major versions are considered completely incompatible, so the likely result is that people would be stuck not able to run their apps at all.  Whereas if people get deprecation warnings and errors, at least they can update their apps and pressure packages to update upstream.

okada:
+1

v3rron:
+1

ffxsam:
Isn't it possible for package authors to pin their package at a specific version? So couldn't you just pin `react` at 0.1.13, so in order for people to update that package, they have to explicitly run `meteor add react@0.1.14`?

stubailo:
@ffxsam that's true!

oles-io:
Would love to see this bumped to version 0.14, along with updated examples after :)

ccorcos:
If you haven't check it out yet `webpack:webpack` is pretty amazing -- you may want to look into using it for React

stubailo:
@ccorcos will that make this package better somehow if we switch to that over cosmos:browserify?

ffxsam:
>If you haven't check it out yet webpack:webpack is pretty amazing -- you may want to look into using it for React

[...]

PR #117, opened 2015-10-15
Upgrade to React 0.14.0

stubailo:
This is a work in progress!

Addresses #116

### Known todo items

- [ ] Figure out how we are supposed to run tests for `react-runtime-prod` since it's `prodOnly`...?
- [ ] Figure out what to do about `ReactDOM` which is now split out from `React`. Apparently there are some backcompat warnings, so maybe it's not super terrible.
- [ ] `React.addons` will no longer be defined, now you need to require it separately. What are the side effects of this?

@dgreensp and @benjamn, I'd appreciate if you could help out with this by looking over this change, the todo items above, and helping identify other issues based on the [change log](https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html).


yyx990803:
Closing since I assume this is no longer relevant.

PR #118, opened 2015-10-16
Upgrade to React 0.14.0.

fabienb4:
Added ReactDOM, ReactDOMServer, React.addons.

Updated docs.

 #117 #116


stubailo:
Looks great - the tests pass! It doesn't give the user fine-grained control over which addons they have, etc, but we can always add that in a later version of the integration.

PR #124, opened 2015-10-30
Fix the typo error which does not expose ReactDOMServer

arunoda:
Right now, React DOM Server is not exposing properly. This will fix it.


arunoda:
@stubailo Seems like we don't have test cases for these.

arunoda:
I've tested this manually and works pretty well. Make sure we've some test cases for these as well. We are using SSR :)

PR #149, opened 2015-12-16
Remove webcomponents.js from includes list.

tmeasday:
This file (from Polymer) includes some ES5 getters which are incompatible with IE8. We include it incase people want to `require('react/lib/webcomponents)`, however I don't think this file is truly a part of react (I think it ends up in the npm module by accident perhaps? not sure).

See https://github.com/meteor/meteor/issues/5791


stubailo:
Looks great. Perhaps we should PR React to not include this in the NPM module? In that commit I found, it looked like they _used_ to have a filter to remove this file.

tmeasday:
Yeah, I reckon we probably should?

PR #151, opened 2015-12-17
Use the Meteor module system instead of browserify.

benjamn:
Fixes https://github.com/meteor/meteor/issues/5795.

These changes can only be used with versions of Meteor that support modules, so I think there need to be some minor version bumps and some `api.versionsFrom("1.3")` statements, but I'm not exactly sure how that works. Help welcome!

Note: because the `react-runtime-{dev,prod}` packages now use `ecmascript-runtime` (implied by `ecmascript`), the `Object.create` and `Object.freeze` shams are no longer necessary.

Browserify provides a global polyfill for `process`, but React only ever uses `process.env.NODE_ENV`, so I was able to provide a minimal polyfill in `react-runtime-{dev,prod}/client-process.js`.

The other `browserify-replace-g` replacements were just rewordings of development-only error messages, which don't seem worth preserving now that we're no longer using browserify.

[This tutorial](http://react-in-meteor.readthedocs.org/en/latest/client-npm/) will need to be rewritten.

cc @avital @stubailo @yyx990803 @martijnwalraven


stubailo:
One possible concern - are we OK with new versions of React only being usable in Meteor >= 1.3? I guess the alternative is doing this manual update process for every version, which isn't great.

packages/react-runtime-dev/package.js
```diff
@@ -26,11 +26,13 @@ Npm.depends({
 });
 
 Package.onUse(function (api) {
-  api.use('cosmos:browserify@0.9.2');
+  api.use('ecmascript');
+
+  api.addFiles('client-process.js', 'client');
+  api.export('process', 'client');
```
stubailo: Do we need this export?

benjamn:
I suspect most/all of our React users will either be excited to upgrade or will at least understand that upgrading is worth the extra effort.

stubailo:
Also, how do people use NPM packages like `material-ui` with this? If `material-ui` does a `require("react/lib/something")` will it get the right module?

stubailo:
Yeah I think I agree - probably most or all React devs have been waiting for modules anyway.

packages/react-runtime-dev/package.js
benjamn: Yes, so that that `process` variable will be declared as a package `var` instead of leaking into the global scope.

packages/react-runtime-prod/client-process.js
```diff
@@ -0,0 +1,5 @@
+process = {
+  env: {
+    NODE_ENV: "development"
```
yyx990803: Should it be "production" here?

yyx990803:
On the other hand - wouldn't it be better if users can just `npm install react react-dom` with the new module system?

benjamn:
> Also, how do people use NPM packages like `material-ui` with this? If `material-ui` does a `require("react/lib/something")` will it get the right module?

That `require` call doesn't work right now, and it's going to be difficult to make it work as long as there's also a Meteor package named `react`. Fortunately, `material-ui` seems to stick to `import React from 'react';` which works fine as long as the Meteor `react` package exports a `React` variable.

I've given some thought to implementing support for Meteor packages installing (some of) their npm dependencies "globally," so maybe that avenue deserves some more investigation.

benjamn:
@yyx990803 yes, in fact I think we could turn all of the Meteor packages in this repository into npm packages, except `jsx`, which we might as well merge into the `ecmascript` package, honestly.

If that simplification works, it could be a big news item for 1.3.

yyx990803:
@benjamn that's what I want to push for in 1.3. The new React experience should be just `npm install` - no more wrapper package / browserify hacks.

Re `jsx`: it seems most React users are using Babel to transpile JSX anyway, so it should be just adding the following babel plugins:

- http://babeljs.io/docs/plugins/transform-react-jsx/
- http://babeljs.io/docs/plugins/syntax-jsx/

Somewhat unrelated: can users configure their own babel plugins when using `ecmascript`? For example they may want to use https://github.com/gaearon/babel-plugin-react-transform

casperanderseneu:
@yyx990803 seem to have understood our needs. no wrapper packages. smooth integration.

To put it into context, what we are looking is preparation for smooth integration with eg. @eXon 's webpack:webpack as 1.3 with modules alone is not going to give us HMR, Code Splitting and access to all the existing webpack loaders.

@eXon is also working on integration with 1.3, so it wouldn't be too bad if you guys communicated.

tmeasday:
This is now much simpler #171

PR #161, opened 2016-01-11
ensure this.container is set in the onDestroyed callback before passing to ReactDOM to unmount it

davidworkman9:
I ran into a case where the ```onRendered``` callback hadn't been called yet, but the ```onDestroyed``` was called first. Currently my workaround is to create a div ```onCreated``` and set ```this.container``` equal to it, but it would be nice if this was fixed in the package by ensuring that ```this.container``` is set before trying to utilize it.

### DIFF
--- a/packages/react-template-helper/react-template-helper.js
+++ b/packages/react-template-helper/react-template-helper.js
@@ -42,7 +42,8 @@ Template.React.onRendered(function () {
 });
 
 Template.React.onDestroyed(function () {
-  ReactDOM.unmountComponentAtNode(this.container);
+  if (this.container)
+    ReactDOM.unmountComponentAtNode(this.container);
 });
 
 // Gets the name of the template inside of which this instance of `{{>


davidworkman9:
@stubailo, any chance this might get merged in?

stubailo:
@tmeasday I can't actually publish this because there's a new JSX that depends on Meteor 1.3 beta - what should we do?

tmeasday:
@stubailo I'm not quite sure I understand the qn but I think the next version of this package should just

a) have `versionsFrom(1.3)`
b) depend on `ecmascript`, not `jsx`
c) not depend on `react-runtime`, instead "peer npm" depend[1] on react

[1] We definitely need some nomenclature for this

stubailo:
Yep. I'm just saying I can't publish this change until Meteor 1.3 is released.

tmeasday:
Oh. But we can publish beta versions of the packages no?

stubailo:
I don't thiiink so? I guess we should investigate the publish script.

tmeasday:
Yeah. I took a look at it... it does... stuff.

I mean now we only have 3 (4?) packages, we can probably just do it by hand (/ducks/)

PR #164, opened 2016-01-14
Allow using component name instead of component

bitomule:
It would be great if we can just use the component name instead of a component reference.Why can't we just pick the component from global? That would remove the need for a template helper returning the component. This small change won't break anything, just adds componentName option. With this change package can be used just with:

```
{{> React componentName="MyComponent" prop1="Hello"}}
```


tmeasday:
I think we want to move away from things being global (see the theme of the 1.3 release), especially in the React world where they don't have to be.

Thanks though!

Issue #169, opened 2016-01-25
What's next for React + Meteor 1.3


yyx990803:
With the ES2015 modules being enabled in Meteor 1.3, we are planning to recommend a new, better workflow for working with React in Meteor. Here's the outline:

## Direct npm installs

Users now are encouraged to install React directly from npm using `npm install react react-dom` (and any additional addons needed) instead of `meteor add react`.
  
#### Benefits of this approach:
  
  - The current `react-runtime` packages require manual updates to stay up-to-date with React’s frequent release schedule. When installing directly from NPM, users can easily select the specific version of React and the specific list of React addons they want to use.
    
  - It is now much easier to work with 3rd party npm packages that expect React to exist as a peer dependency and directly `require(‘react’)`  in their code. Previously we had to use a Browserify-based hack to replace these.

#### Potential issues:
  
  - Backwards compatibility with old packages or apps that depend on `react-runtime`. Luckily as of release [`1.3-modules-beta.5`](https://github.com/meteor/meteor/issues/5788#issuecomment-174621057), packages can now directly import top-level `node_modules` in an app. The new `react-runtime` package would simply be setting the global `React` by using the npm installed version. If React is not installed from npm, there would be a warning prompting you to do so.

## Meteor Data Integration

The community has been looking for a new pattern to use instead of the mixin approach used by `react-meteor-data`, especially now that ES2015 is the default language, we don't want Meteor's default data integration to block users from using the class-based React component syntax.

There is a lot of community innovation going on in this space, e.g. Arunoda’s work on Mantra. [MDG is also working on major improvements to Meteor's data layer](http://info.meteor.com/blog/reactive-graphql). At this very moment, the absolute best practice is yet to be defined. So for now what we are going to provide is a minimal addition to the current `react-meteor-data` package, essentially a higher-order-component based on the current `ReactMeteorData` mixin:

``` js
import React from 'react';

ReactMeteorDataContainer = function (Component, getMeteorData) {
  return React.createClass({
    mixins: [ReactMeteorData],
    getMeteorData() {
      return getMeteorData(this.props);
    },
    render() {
      return <Component {...this.props} {...this.data} />
    }
  });
};
```

And with modules, an example component using this function:

``` js
import React from 'react';
import { Meteor } from 'meteor/meteor';
import { Todos } from '../lib/collections';
import { ReactMeteorDataContainer } from 'meteor/react-meteor-data';

// The app component is "pure" and only depend on props
export class App extends React.Component {
  render() {
[...]

PR #172, opened 2016-03-05
Added new `createContainer` API.

tmeasday:
I've updated the `react-meteor-data` package to provide the `createContainer` API as discussed elsewhere (notably https://github.com/meteor/react-packages/issues/169).

@stubailo I'm assigning this to you again; apologies ;)

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -1,4 +1,4 @@
-ReactMeteorData = {
+const ReactMeteorData = {
   componentWillMount() {
     this.data = {};
     this._meteorDataManager = new MeteorDataManager(this);
@@ -29,7 +29,7 @@ ReactMeteorData = {
   },
   componentWillUnmount() {
     this._meteorDataManager.dispose();
-  }
+  },
 };
 
 // A class to keep the state and utility methods needed to manage
@@ -147,3 +147,5 @@ class MeteorDataManager {
     this.oldData = newData;
   }
 }
+
+export default ReactMeteorData;
--- a/packages/react-meteor-data/createContainer.jsx
+++ b/packages/react-meteor-data/createContainer.jsx
@@ -1,0 +1,39 @@
+/**
+ * Container helper using react-meteor-data.
+ */
+
+import React from 'react';
+import PureRenderMixin from 'react-addons-pure-render-mixin';
+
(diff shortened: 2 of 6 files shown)


tmeasday:
Note that I'd consider the old `MeteorDataMixin` as being deprecated in this version.

stubailo:
Is it weird to use the mixin in the implementation?

tmeasday:
Do you think? I mean the mixin is battle-tested (to some degree) and has a lot of code to deal with various edge cases. I don't think there's any strong reason *not* to -- except that it was suggested that we should use komposer behind the scenes. I'm not convinced it has the same level of usage behind it though.

stubailo:
That's a fair point. We should look at any issues with the mixin though, I figured maybe we didn't need to if we were moving away.

genyded:
Hey Sashko and Tom,

Given the general debate and uncertainty around mixins in general and the traction komposer/mantra seem to be gaining (for now), might it make sense to consider a flag or something and allowing either/both to be used (maybe one as experimental initially)? That way no matter which way the wind ends up blowing, switching or finalizing should be as simple as toggling a flag.

I personally think data fetching is yet another 'battle' (like view layers, containers, routers, build systems, etc...) which doesn't need 6 available solutions. That usually just confuses folks on which one to use and why or leads to countless amounts of time trying to self ascertain the benefits or differences. That approach does however seem to be becoming the norm in the JS community as a whole. Every time someone comes out with something. 5 others have to come out with their version of pretty much the same thing. I'm sure that is as much an ongoing challenge for MDG as it is the rest of us. 

However initially providing two solutions to a piece of the pie that are easily switchable seems like it may help drive narrowing the choices while sill providing options - which in turn may help attract more developers all while helping grow/mature community solutions in the process.

The data fetching scenario might be a great candidate to try something like this because generally speaking it is much less complex and has less code overall than a lot of other areas. 

Just some food for thought.

stubailo:
> might it make sense to consider a flag or something and allowing either/both to be used (maybe one as experimental initially)?

The Meteor 1.3 version of `react-meteor-data` supports both approaches - a mixin and a container component. The container component will be the recommended approach.

stubailo:
@tmeasday is this something we are merging soon?

tmeasday:
Yeah this has been merged into devel and release as `0.2.6-beta.16`

PR #182, opened 2016-03-25
Update community.md

perak:
Meteor Kitchen now can generate Meteor+React app, so I think it deserves to be listed here. :)

### DIFF
--- a/docs/community.md
+++ b/docs/community.md
@@ -7,3 +7,4 @@ These tools work with the packages explained in this guide, and give you extra f
 3. [Arunoda's blog post about server-side rendering](https://kadira.io/blog/meteor-ssr-support-using-flow-router-and-react/)
 4. Material UI: [izzilab:material-ui](https://atmospherejs.com/izzilab/material-ui)
 5. Use Blaze Templates as / in React Components: [4commerce:blaze-as-react](https://atmospherejs.com/4commerce/blaze-as-react)
+6. Meteor+React code generator: [Meteor Kitchen](http://www.meteorkitchen.com)


stubailo:
Hmm, this was back in the day where the list of community packages that supported react was so small they could be listed on a tiny page like this. In the post-Meteor 1.3 world, there are now hundreds of packages and tools that work with React.

Also we are removing this whole guide in favor of: http://guide.meteor.com/v1.3/react.html and the rest of the guide which also tells you how to use NPM, etc.

Do you have any idea where it would be best to feature packages like this?

perak:
@stubailo OK, I understand. 

But... to bad to remove this whole guide - Meteor's official guide is good for beginners, but not useful for more advanced users at all. This guide helped me a lot to move from Blaze to React (plus using 5-10 different blogs and source codes, all of which showing different techniques to work with Meteor+React... and my transition was (or still is) more painful than it should be.

Full respect to MDG efforts, but I'm not satisfied with official docs and guides. I understand philosophy to gave users a freedom of choice, but one detailed guide for "one of possible ways" to use Meteor is missing.

*Do you have any idea where it would be best to feature packages like this?*
Sorry, I don't understand - packages like which?

stubailo:
> Meteor's official guide is good for beginners, but not useful for more advanced users at all. This guide helped me a lot to move from Blaze to React (plus using 5-10 different blogs and source codes, all of which showing different techniques to work with Meteor+React... and my transition was (or still is) more painful than it should be.

Do you think the official guide is not a good replacement for this React Guide? if so, we should improve the Meteor Guide to be just as useful. @tmeasday something to think about.

> one detailed guide for "one of possible ways" to use Meteor is missing.

Is guide.meteor.com not enough for this use case?

perak:
Excuse me, I wish to say: one detailed guide for "one of possible ways" to use **Meteor+React** is missing.

Current meteor guide is not bad (I wish it was as good as today in times when I was learning Meteor ~2.5 yrs ago.

perak:
BTW, most of "meteor kitchen" users are actually using it to learn Meteor: they describe and generate app, and then they use generated code to understand how to solve some everyday programming problems.

But, code generated with meteor-kitchen is not "best practices". Most of code is outdated or was never "best practice" because I was not sure what is best practice in time when I was implementing it. (and to be honest, I was too lazy to fix the things later :) ).

Now I decided to make it better with Meteor 1.3+React, but I'm again in the same position - what is damn best practice? (for lot of things). 

Maybe I started too early - before things are crystallised...

stubailo:
> what is damn best practice? (for lot of things).

Let's talk about it and put it in the guide! I think it's basically up to us and the rest of the community to make it happen.

perak:
Sure, let's do it!

perak:
Is this right place for talk, or I can contact you somewhere else?

stubailo:
Please contact me on the forums or file an issue on the guide!

I've removed this guide entirely, and put links to the old content and the relevant sections of the guide.

perak:
OK, noted.

PR #188, opened 2016-04-19
Don't update data in place, instead create a new object

stubailo:
It seems like the old implementation cared about other things using `this.data` but I don't think that's reasonable anymore. On the other hand, keeping the `this.data` object with the same reference actively breaks Redux because it relies on comparing prop references to detect changes.


stubailo:
Wait, I don't think this makes sense.

PR #197, opened 2016-07-28
Temporarily assign this.context as well

mart-jansink:
Previously the getMeteorData function would see the old context. This is unexpected because the props and state appear to be already updated. So just temporarily assign this.context just like it's done with this.props and this.state.

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -5,13 +5,14 @@ const ReactMeteorData = {
     const newData = this._meteorDataManager.calculateData();
     this._meteorDataManager.updateData(newData);
   },
-  componentWillUpdate(nextProps, nextState) {
+  componentWillUpdate(nextProps, nextState, nextContext) {
     const saveProps = this.props;
     const saveState = this.state;
+    const saveContext = this.context;
     let newData;
     try {
-      // Temporarily assign this.state and this.props,
-      // so that they are seen by getMeteorData!
+      // Temporarily assign this.state, this.props and
+      // this.context, so that they are seen by getMeteorData!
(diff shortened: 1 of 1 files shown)


Twisterking:
Is there a way to access `this.context` inside the `createContainer` function? I would need this desperately but it doesn't work!

dburles:
Hey @Twisterking what's the use-case there?

Twisterking:
Well in general there are sometimes variables which I do not want to pass down to each child component via props but just set it via context once and then having it available in all children by default. I might need to access this context vars in my Meteor data container then and subscribe to any publications properly. Right now I always have to make a work around and pass this context var into a prop and then use it in the `createContainer` function. I would be much more convenient if there would be any way to access the context directly.

dburles:
I'm almost certain that there's a [better approach to that than using context](https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076), so I don't really think there's much value in adding it. Your workaround seems okay in the rare instance that someone may need to reference context within `withTracker`/`createContainer`.

mariorodriguespt:
@dburles In my use case, getting access to the `context` inside a data container is a must. I'm dealing with subdomains and I'm using the context to propagate global fields to the components. Since a data container needs those fields in order to subscribe, makes total sense to allow the context inside `withTracker`.

I would like to create a pull for this because this is already working on my local machine. The modification is simply and backwards compatible: https://gist.github.com/mariorodriguespt/6a1efa2ab8ae2ceaee79d6ec773e5fd7

Let me know if I'm missing any important detail, if not I'll create a PR with this.
The usage would become something like this, again, backwards compatible: https://gist.github.com/mariorodriguespt/e873d94ba59c688bf6071f468b9800c2

filipenevola:

PR #205, opened 2016-10-29
export createContainer (in the package.js)

Bramas:
after doing `console.log(require('meteor/react-meteor-data'));` I realized that createContainer is not exported by the package. And, it's normal because in the package.js, only `ReactMeteorData` is exported.

So even the trivial example starting with `import { createContainer } from 'meteor/react-meteor-data';` fails.

I just added `createContainer` in the export like it should be.


dburles:
I don't believe this is still an issue, @Bramas ?

Bramas:
Sorry but I don't know (I think you can close it if there's no issue related to that)

hwillson:
Closing as this should no longer be an issue. Thanks!

Issue #203, opened 2016-10-08
createContainer that allows creating ReactiveVar in its scope


raaymax:
It would be nice to have ability to create ReactiveVar somewhere in scope of container. Creating wrapper object for wrapper object just for variable declaration is a little redundant.
Here is example usage:

``` js
export default createContainer({
    getInitialMeteorData(){
        return {
            selection: new ReactiveVar(null)
        };
    },
    getMeteorData(props, initial){
        var imageId = initial.selection.get();
        var avatar = null;
        if(avatarId) {
            avatar = Files.findOne({_id: imageId});
        }
        return {
            selection: initial.selection,
            avatar: avatar,
        };
    }
}, UserForm);
```

What do you think?

ecwyne:
@raaymax how are you able to achieve this currently? I'm running into this exact situation.

mokolodi1:
See #213!

@raaymax I've opted to use the React-compatible name `getInitialState` and then I pass that state into `getMeteorData` as an argument as well as via `this.state`.

filipenevola:
I'm closing this just because it's too old. We can open new issues for items that are still valid.

Most of these items were solved already or replaced by new strategies (like hooks)

PR #213, opened 2017-03-14
Add getInitialState option to createContainer

mokolodi1:
This addresses #203 and adds an option to give a `getInitialState` function to `createContainer`. 

As a former Blaze user, I can't understand how people use React without `ReactiveVar`s, and this PR makes it possible to use them with the best-practice method `createContainer`.


dburles:
Hey @mokolodi1 unless I'm missing something here, the proposed changes seem a little unnecessary. Why not simply define your ReactiveVar's like so:

```js
const loadCount = new ReactiveVar(props.initialLoadCount);

export default createContainer(() => ({
  loadCount: loadCount.get(),
  loadMore: () => loadCount.set(loadCount.get() + 5),
  // ... etc
}), Foo);
```

mokolodi1:
I agree it doesn't seem to be the _best_ API per se. `getInitialState` was the only place I could think of to attach the closure.

filipenevola:

PR #217, opened 2017-04-10
Convert ReactMeteorData to a Higher-Order Component

### LINKED ISSUE #200: Remove "react-addons-pure-render-mixin"?
I have a question. In regards to this post, [https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html#solution](https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html#solution) would it be possible to remove "react-addons-pure-render-mixin" from "react-meteor-data"? They suggest to start using `var shallowCompare = require('react-addons-shallow-compare');` for `shouldComponentUpdate` I am not sure if this is possible or not? Thanks

### LINKED ISSUE #216: Updating react-addons-pure-render-mixin to React.PureComponent
React 15.5 is discontinuing support for React add-ons, including pure-render-mixin, as mentioned in their release notes: https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

Plans to update meteor-react-data?

leocavalcante:
Converts the ReactMeteorData mixin to a [Higher-Order Component](https://facebook.github.io/react/docs/higher-order-components.html)

This PR fixes warnings with latest React like:
* React.createClass is deprecated and will be removed in version 16.

Fixes #216
Closes #200 
(maybe more)

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -1,53 +1,53 @@
 import React from 'react';
 
-export function connect({ getMeteorData, pure = true }, WrappedComponent) {
+export function connect({ getMeteorData, pure = true }) {
   const BaseComponent = pure ? React.PureComponent : React.Component;
 
-  class ReactMeteorData extends BaseComponent {
-    getMeteorData() {
-      return getMeteorData(this.props);
-    }
-
-    componentWillMount() {
-      this.data = {};
-      this._meteorDataManager = new MeteorDataManager(this);
-      const newData = this._meteorDataManager.calculateData();
-      this._meteorDataManager.updateData(newData);
-    }
-
-    componentWillUpdate(nextProps, nextState) {
-      const saveProps = this.props;
-      const saveState = this.state;
-      let newData;
-      try {
-        // Temporarily assign this.state and this.props,
-        // so that they are seen by getMeteorData!
-        // This is a simulation of how the proposed Observe API
-        // for React will work, which calls observe() after
-        // componentWillUpdate and after props and state are
-        // updated, but before render() is called.
-        // See https://github.com/facebook/react/issues/3398.
-        this.props = nextProps;
-        this.state = nextState;
-        newData = this._meteorDataManager.calculateData();
-      } finally {
-        this.props = saveProps;
-        this.state = saveState;
+  return WrappedComponent => (
+    class ReactMeteorData extends BaseComponent {
+      getMeteorData() {
+        return getMeteorData(this.props);
       }
 
-      this._meteorDataManager.updateData(newData);
-    }
+      componentWillMount() {
+        this.data = {};
+        this._meteorDataManager = new MeteorDataManager(this);
+        const newData = this._meteorDataManager.calculateData();
+        this._meteorDataManager.updateData(newData);
+      }
 
-    componentWillUnmount() {
-      this._meteorDataManager.dispose();
-    }
+      componentWillUpdate(nextProps, nextState) {
+        const saveProps = this.props;
+        const saveState = this.state;
+        let newData;
+        try {
+          // Temporarily assign this.state and this.props,
+          // so that they are seen by getMeteorData!
+          // This is a simulation of how the proposed Observe API
+          // for React will work, which calls observe() after
+          // componentWillUpdate and after props and state are
+          // updated, but before render() is called.
+          // See https://github.com/facebook/react/issues/3398.
+          this.props = nextProps;
+          this.state = nextState;
+          newData = this._meteorDataManager.calculateData();
+        } finally {
+          this.props = saveProps;
+          this.state = saveState;
+        }
 
-    render() {
-      return <WrappedComponent {...this.props} {...this.data} />;
-    }
-  }
+        this._meteorDataManager.updateData(newData);
+      }
 
-  return ReactMeteorData;
+      componentWillUnmount() {
+        this._meteorDataManager.dispose();
+      }
+
+      render() {
+        return <WrappedComponent {...this.props} {...this.data} />;
+      }
+    }
+  );
 }
 
 // A class to keep the state and utility methods needed to manage
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -1,36 +1,52 @@
-const ReactMeteorData = {
-  componentWillMount() {
-    this.data = {};
-    this._meteorDataManager = new MeteorDataManager(this);
-    const newData = this._meteorDataManager.calculateData();
-    this._meteorDataManager.updateData(newData);
-  },
-  componentWillUpdate(nextProps, nextState) {
-    const saveProps = this.props;
-    const saveState = this.state;
-    let newData;
-    try {
-      // Temporarily assign this.state and this.props,
-      // so that they are seen by getMeteorData!
-      // This is a simulation of how the proposed Observe API
-      // for React will work, which calls observe() after
-      // componentWillUpdate and after props and state are
-      // updated, but before render() is called.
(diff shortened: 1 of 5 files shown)


packages/react-meteor-data/ReactMeteorData.jsx
```diff
@@ -1,36 +1,52 @@
...
-      this.props = nextProps;
-      this.state = nextState;
-      newData = this._meteorDataManager.calculateData();
-    } finally {
-      this.props = saveProps;
-      this.state = saveState;
+import React from 'react';
+
+export function connect({ getMeteorData }, WrappedComponent) {
+  class ReactMeteorData extends React.Component {
```
s-ol: shouldn't this be a `React.PureComponent`?

menelike: ... and sometimes it should not be pure at all, this is why this is optional and set to true per default https://github.com/meteor/react-packages/blob/2e0cbe94b1a8d657559e223006ed0507b212829c/packages/react-meteor-data/createContainer.jsx#L20

leocavalcante: Working on it.

leocavalcante:
My test app, since there is no unit tests, was:

```js
import React, { Component } from 'react'
import { render } from 'react-dom'
import { Meteor } from 'meteor/meteor'
import { Mongo } from 'meteor/mongo'
import { createContainer } from 'meteor/react-meteor-data'

class MessageList extends Component {
  render() {
    const { messages } = this.props

    console.log(messages)

    return (
      <ul>
        {messages.map(message => <li>{message.body}</li>)}
      </ul>
    )
  }
}

const Messages = new Mongo.Collection('messages')

if (Meteor.isServer) {
  Meteor.publish('messages', () => {
    return Messages.find({})
  })
}

if (Meteor.isClient) {
  const Container = createContainer(() => {
    const handle = Meteor.subscribe('messages')
    const messages = Messages.find({}).fetch()

    return {
      ready: handle.ready(),
      messages
    }
  }, MessageList)

  Meteor.startup(() => render(<Container />, document.getElementById('root')))
}
```

And I always get impressed how easy is to build a real-time database-backed app with Meteor :open_mouth:

packages/react-meteor-data/ReactMeteorData.jsx
```diff
@@ -1,36 +1,54 @@
...
-      // See https://github.com/facebook/react/issues/3398.
-      this.props = nextProps;
-      this.state = nextState;
-      newData = this._meteorDataManager.calculateData();
-    } finally {
-      this.props = saveProps;
-      this.state = saveState;
+import React from 'react';
+
+export function connect({ getMeteorData, pure = true }, WrappedComponent) {
```
leocavalcante: Hi @benjamn, thanks for review. There is no particular reason. Actually, the [convention for `connect`](https://facebook.github.io/react/docs/higher-order-components.html#convention-maximizing-composability) is to be a high-order function that returns a high-component. Like Redux's `connect` and Apollo's `graphql`. You think we should go by that?

dburles: Related https://github.com/meteor/react-packages/pull/215

hwillson:
Hi @leocavalcante - would you mind rebasing this PR? As a temporary fix we just merged PR #218, to help get rid of the `React.createClass` deprecation warnings. This PR represents a better long term solution however, so if you're able to rebase, I'll start doing some thorough testing of these changes. Thanks!

leocavalcante:
Hi @hwillson - here is it.

packages/react-meteor-data/react-meteor-data.jsx
```diff
@@ -1,11 +1,9 @@
...
   react: '15.x',
-  'react-addons-pure-render-mixin': '15.x',
-  'create-react-class': '15.x',
 }, 'react-meteor-data');
 
 const createContainer = require('./createContainer.jsx').default;
-const ReactMeteorData = require('./ReactMeteorData.jsx').default;
 
-export { createContainer, ReactMeteorData };
+export { createContainer };
```
benjamn: You should be able to do the remaining export on one line:
```js
export { default as createContainer } from './createContainer.jsx';
```

packages/react-meteor-data/react-meteor-data.jsx
```diff
@@ -1,11 +1,9 @@
 import { checkNpmVersions } from 'meteor/tmeasday:check-npm-versions';
+
 checkNpmVersions({
   react: '15.x',
-  'react-addons-pure-render-mixin': '15.x',
-  'create-react-class': '15.x',
```
benjamn: Why not keep checking `create-react-class` here?

leocavalcante: It is not used anymore.

fabien-h:
When we `meteor add ...` We still have the 0.2.9 and not the 0.2.10.

    meteor add react-meteor-data@0.2.10
    error: no such version react-meteor-data@0.2.10

hwillson:
Hi @fabien-h - version 0.2.10 hasn't been published yet. The most recent published version is [0.2.9](https://atmospherejs.com/meteor/react-meteor-data). A new (published) version is coming soon though - thanks!

abernix:
If anyone wants to try this right now, I've published `react-meteor-data@0.2.11-beta.1` which includes this, on top of #218 (which is essentially nullified, but worth saving as a worth contribution by @merlinpatt).

Feedback appreciated!

rijk:
I have a weird issue with this. Changes in context don't seem to propagate through this component.. The context is there, but an update doesn't come through when it has to pass down the tree through a `ReactMeteorData` container. They do through normal components.

rijk:
Figured it out, it works when I use the `pure: false` option.

mhagmajer:
Doesn't this break the default export of https://github.com/meteor/react-packages/blob/devel/packages/react-meteor-data/ReactMeteorData.jsx ?

In our project we're using this in the following way:

```javascript
import { ReactMeteorData } from 'meteor/react-meteor-data';
import React from 'react';

export default class MeteorComponent<D, P, S, M> extends React.Component<D, P, S> {
  static defaultProps: D;
  props: P;
  state: S;
  data: M;

  getMeteorData(): M {
    return ({}: any);
  }

  componentWillMount() {
    ReactMeteorData.componentWillMount.bind(this)();
  }

  componentWillUpdate(nextProps: P, nextState: S) {
    ReactMeteorData.componentWillUpdate.bind(this)(nextProps, nextState);
  }

  componentWillUnmount() {
    ReactMeteorData.componentWillUnmount.bind(this)();
  }
}
```

And `MeteorComponent` is the base class for all our react component that can track changes to Meteor data.

leocavalcante:
You are right. Sorry about that. This change is actually a BC breaker since `ReactMeteorData` isn't a plain object anymore and the package didn't export it.

As workaround, you should be able to get those methods back as a dummy container prototype:
```js
import { createContainer } from 'meteor/react-meteor-data'

const MeteorContainer = createContainer(() => { }, () => { })
const ReactMeteorData = MeteorContainer.prototype

console.log(ReactMeteorData.componentWillMount)
```

To @benjamn @hwillson and @abernix 
I'm not in the position to make a call here, but I'll be glad to make a PR to fix whatever you decide 😃

hwillson:
Thanks for the heads up @mhagmajer. We've gone ahead and published the new changes since they were being asked for by a large number of users. As you've mentioned though, the new version does break backwards compatibility in some cases. To make sure we don't lose track of this, would you mind creating a new issue for this, outlining exactly what you've commented here? We can then continue the discussion in that new issue thread, and reach out to others for feedback (as well as gather 👍's to see how many people are affected by this). Sorry for any inconvenience this has caused. For now you can consider version pinning at 0.2.9. Thanks!

hwillson:
@mhagmajer PR #224 should help resolve your issue. Hopefully we can get it merged/published soon. Thanks!

PR #218, opened 2017-04-24
Switch to create-react-class

merlinpatt:
React 15.5 has deprecated React.createClass. This fixes that. Prompted by [this issue](https://github.com/meteor/meteor/issues/8625)

### DIFF
--- a/packages/react-meteor-data/createContainer.jsx
+++ b/packages/react-meteor-data/createContainer.jsx
@@ -2,7 +2,7 @@
  * Container helper using react-meteor-data.
  */
 
-import createClass from 'create-react-class';
+import createReactClass from 'create-react-class';
 import React from 'react';
 import PureRenderMixin from 'react-addons-pure-render-mixin';
 
@@ -27,7 +27,7 @@ export default function createContainer(options = {}, Component) {
   }
 
   /* eslint-disable react/prefer-es6-class */
-  return createClass({
+  return createReactClass({
     displayName: 'MeteorDataContainer',
     mixins,
     getMeteorData() {
--- a/packages/react-meteor-data/createContainer.jsx
+++ b/packages/react-meteor-data/createContainer.jsx
@@ -2,6 +2,7 @@
  * Container helper using react-meteor-data.
  */
 
+import createClass from 'create-react-class';
 import React from 'react';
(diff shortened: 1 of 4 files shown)


merlinpatt:
This offers a quick fix to the issue given above. It might be better to instead merge in the [higher order component PR](https://github.com/meteor/react-packages/pull/217)

packages/react-meteor-data/createContainer.jsx
```diff
@@ -2,6 +2,7 @@
  * Container helper using react-meteor-data.
  */
 
+import createClass from 'create-react-class';
```
hwillson: To make sure we stay in-line with the [React docs](https://facebook.github.io/react/docs/react-without-es6.html), can we change `createClass` to `createReactClass`?

packages/react-meteor-data/createContainer.jsx
```diff
@@ -26,7 +27,7 @@ export default function createContainer(options = {}, Component) {
   }
 
   /* eslint-disable react/prefer-es6-class */
-  return React.createClass({
+  return createClass({
```
hwillson: Same comment as above - we might want to change `createClass` to `createReactClass`.

merlinpatt:
@hwillson updated per your changes

hwillson:
Looks great @merlinpatt - LGTM!

abernix:
Any other docs/tutorials/guide/etc. that need to be updated with this change?

hwillson:
Good question @abernix - I guess the [Using createContainer](https://guide.meteor.com/react.html#using-createContainer) section of the Guide could be updated to mention that after adding the `react-meteor-data` package, people will also have to `meteor npm install --save create-react-class`. But then again, `checkNpmVersions` will automatically tell them that, and we're also requiring `react-addons-pure-render-mixin` in the same way (and not mentioning it in the Guide). So maybe we're okay?

Other than the above, I don't think there is much we need to say about the cutover to using `create-react-class` internally, since its use should be transparent to users (other than requiring the `create-react-class` package be installed).

hwillson:
Oh, but just to add - once published I'm going to update the react branch of `todos` to use this version (for reference).

PR #224, opened 2017-05-26
Restored the ReactMeteorData object so it can still be used with mixins.

### LINKED ISSUE #223: react-meteor-data v0.2.11 breaks backwards compatibility
In the latest 0.2.11: the global ReactMeteorData variable is exported as 'undefined'.
Which breaks old code relying on the global ReactMeteorData mixin and React.createClass with getMeteorData().

hwillson:
This PR is intended to help resolve backwards compatibility issues that were introduced in PR #217, and will fix #223. The newer HOC approach is maintained with these changes, but the pre-existing `ReactMeteorData` object is preserved and exported, which allows people using mixin approaches to continue using the latest version of this package. 

We should probably consider releasing a beta version with these changes.

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -1,54 +1,8 @@
+/* global Package */
+
 import React from 'react';
-
-export function connect({ getMeteorData, pure = true }) {
-  const BaseComponent = pure ? React.PureComponent : React.Component;
-
-  return WrappedComponent => (
-    class ReactMeteorData extends BaseComponent {
-      getMeteorData() {
-        return getMeteorData(this.props);
-      }
-
-      componentWillMount() {
-        this.data = {};
-        this._meteorDataManager = new MeteorDataManager(this);
-        const newData = this._meteorDataManager.calculateData();
-        this._meteorDataManager.updateData(newData);
-      }
-
-      componentWillUpdate(nextProps, nextState) {
-        const saveProps = this.props;
-        const saveState = this.state;
-        let newData;
-        try {
-          // Temporarily assign this.state and this.props,
-          // so that they are seen by getMeteorData!
-          // This is a simulation of how the proposed Observe API
-          // for React will work, which calls observe() after
-          // componentWillUpdate and after props and state are
-          // updated, but before render() is called.
-          // See https://github.com/facebook/react/issues/3398.
-          this.props = nextProps;
-          this.state = nextState;
-          newData = this._meteorDataManager.calculateData();
-        } finally {
-          this.props = saveProps;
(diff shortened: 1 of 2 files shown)


polkhovsky:
Any idea when this will be pulled? This is a major showstopper.

packages/react-meteor-data/ReactMeteorData.jsx
```diff
@@ -166,4 +122,61 @@ class MeteorDataManager {
...
+    class ReactMeteorDataComponent extends BaseComponent {
+      getMeteorData() {
+        return getMeteorData(this.props);
+      }
+      render() {
+        return <WrappedComponent {...this.props} {...this.data} />;
+      }
+    }
+
+    ReactMeteorDataComponent.prototype.componentWillMount =
```
abernix: Instead of assigning in this way, couldn't these `componentWill{Mount,Update,Unmount}` methods represent an intermediary object which was created (first) by `extend`ing onto `BaseComponent` and then have `getMeteorData` and `render` further `extend`ed onto that intermediary object to get the resulting `ReactMeteorDataComponent`?

packages/react-meteor-data/ReactMeteorData.jsx
```diff
@@ -166,4 +122,61 @@ class MeteorDataManager {
...
+    this._meteorDataManager.updateData(newData);
+  },
+
+  componentWillUnmount() {
+    this._meteorDataManager.dispose();
+  },
+};
+
+export function connect({ getMeteorData, pure = true }) {
+  const BaseComponent = pure ? React.PureComponent : React.Component;
```
abernix: (Read this comment after my below comment) :wink:

Since this may be called very often (for each of a potentially large number of components, no?), might there be a performance gain by defining two intermediary "base" objects (one pure and one non-pure) before the `connect` method and then just doing the single `extend` within the `WrappedComponent` function?

hwillson: Adjusting (see next comment) - thanks!

packages/react-meteor-data/ReactMeteorData.jsx
hwillson: Thanks @abernix - we can certainly get close to this. I can leave the `ReactMeteorData` object as is, create 2 new classes called `ReactComponent` and `ReactPureComponent` (that extend `React.Component` and `React.PureComponent` respectively), then copy all `ReactMeteorData` properties into the new classes. We can then use those new classes within the `connect` function. I'll adjust - thanks!

abernix:
Beta published as `meteor-react-data@0.2.12-beta.0`!

abernix:
Re-published in `meteor-react-data@0.2.12-beta.1`.

PR #226, opened 2017-07-10
Remove react-addons-test-utils and use react-dom/test-utils

pletcher:
This change fixes a deprecation warning present in every Meteor app using React.

At the moment, the error that results amounts to console noise, but it's easy to imagine `react-dom/test-utils` adding functionality while `react-addons-test-utils` will only stagnate.

I don't think this change would cause any problems for apps that depend on this package, but I'm happy to add safeguards if needed.

### DIFF
--- a/packages/react-runtime/react-runtime.js
+++ b/packages/react-runtime/react-runtime.js
@@ -14,7 +14,6 @@ const requiredPackages = {
 
 if (Meteor.isDevelopment) {
   _.extend(requiredPackages, {
-    'react-addons-test-utils': '15.x',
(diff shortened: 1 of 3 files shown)


robertpitt:
Any updates on this being merged, this error is very annoying ?

dburles:
Hey @pletcher @robertpitt I've not actually seen this error. Is it still an issue on your end?

pletcher:
Yep, I'm still getting it. Out of curiosity, what version of React are you using? I might have just set things up in an un-Meteor-y way :)

filipenevola:

PR #232, opened 2017-09-28
Update version range for React 15 & 16

klaussner:
This PR updates the `checkNpmVersions` range for React to fix the

```
WARNING: npm peer requirements (for react-meteor-data) not installed:
 - react@16.0.0 installed, react@15.x needed
```

warning if React 16 is installed. I also updated the version range for React 15 (from 15.x to ^15.3) because 15.3.0 is the oldest version that works with `react-meteor-data`.

Fixes #214

PR #238, opened 2017-10-19
Update version range to include React 16

aermler:
Updated file similar to #232, haven't tested it.


egemon:
Thanks for that. I need it as well!

paulincai:
Hi, same here. I hope this will be looked at @abernix . It affects mixed Blaze/React projects.

abernix:
Thanks very much for addressing this, @aermler.  I've published this change with `react-template-helper@0.2.10`.  Let me know if you experience any problems with it!

PR #245, opened 2018-01-15
Adds a method to ReactMeteorDataComponent for retrieving the wrapped …

rooch84:
…component.

This is useful if the parent component needs to call a method on the child.

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -180,6 +180,9 @@ export default function connect(options) {
       getMeteorData() {
         return getMeteorData(this.props);
       }
+			getWrappedComponent() {
+			  return WrappedComponent;
+			}
       render() {
         return <WrappedComponent {...this.props} {...this.data} />;
       }


packages/react-meteor-data/ReactMeteorData.jsx
```diff
@@ -180,6 +180,9 @@ export default function connect(options) {
       getMeteorData() {
         return getMeteorData(this.props);
       }
+			getWrappedComponent() {
```
hwillson: Would you mind fixing the spacing here to align with the style used in the rest of the file?

yched:
This would be fixed by #266 I guess.

filipenevola:

derwaldgeist:
Has this been implemented?

yched:
@derwaldgeist in the current release, withTracker() uses React.forwardRef, so you should be able to get the wrapped component symply by using a regular React ref

derwaldgeist:
@yched Thanks for your feedback. Yeah, I realized this in the meantime. I was still using version 1.0.0 of this package so far, where it did not work. Had to update Meteor to 1.10 though, because of version dependencies on the ecmascript package.

Issue #221, opened 2017-05-05
Repo health check


hwillson:
While prepping for the release of PR #217, we've done an overall health review of the full `react-packages` repo. A few quick notes:

- Now that Meteor has full npm support, the [`react`](https://github.com/meteor/react-packages/tree/devel/packages/react) and [`react-runtime`](https://github.com/meteor/react-packages/tree/devel/packages/react-runtime) packages are no longer necessary, and can be deprecated.
- While both the [`react-meteor-data`](https://github.com/meteor/react-packages/tree/devel/packages/react-meteor-data) and [`react-template-helper`](https://github.com/meteor/react-packages/tree/devel/packages/react-template-helper) packages are currently valid and in use, `react-meteor-data` is by far the more popular package, and will likely see updates more frequently than `react-template-helper`. The current `react-packages` publishing process only covers publishing all packages in this repo at the same time however. This was done to make sure version numbers were all kept in synch, but now that `react` is handled via npm, this appears to be no longer necessary. The publishing process should be revisited, now that it can be simplified.
- All package test apps needs to be fixed/updated. They're using older versions of Meteor, and don't appear to be working properly. Package documentation should also be updated to outline how to use the tests (since they're run as standalone Meteor apps).
- Actually, now that there are really only two packages left in this repo that are active, and they don't depend on each other (or other packages from this repo in any way), we might want to consider splitting these packages out into their own individual repos.

filipenevola:
I'm closing this just because it's too old. We can open new issues for items that are still valid.

Most of these items were solved already or replaced by new strategies (like hooks)

PR #249, opened 2018-02-13
add react-ssr-csr package

pravdomil:
this is the missing piece for https://github.com/meteor/meteor/blob/devel/Roadmap.md#out-of-the-box-support-for-advanced-react-features

now you can specify your root component and package does the rest

original repo https://github.com/pravdomil/Meteor-React-SSR-and-CSR-with-loadable-and-subscriptions


hwillson:
@pravdomil Is there any specific reason why this should be managed here, instead of as a community package (like you were doing)? The number of eyes and maintainers on this repo is quite low, so adding a new package here isn't something that is considered lightly. If anything, we'd like to remove packages (see https://github.com/meteor/react-packages/issues/221).

pravdomil:
ok

PR #257, opened 2018-07-05
 Use Tracker.nonreactive when calling lifecycle methods of a React component from within a Tracker computation.

mattmccutchen:
This change fixes a problem seen in my Meteor app.  I'll be happy to provide a full reproduction on request, but I don't want to go to the trouble if the meteor/react-packages repository is as dead as it appears.

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -76,7 +76,12 @@ class MeteorDataManager {
           c.stop();
           // Calling forceUpdate() triggers componentWillUpdate which
           // recalculates getMeteorData() and re-renders the component.
-          component.forceUpdate();
+          //
+          // Use Tracker.nonreactive so that autoruns, subscriptions, etc.
+          // created by the component don't get stopped when this computation is
+          // stopped, because the component is responsible for its own lifecycle
+          // and won't expect to have to recreate them.
+          Tracker.nonreactive(() => component.forceUpdate());
         }
       })
     ));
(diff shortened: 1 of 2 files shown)


mattmccutchen:
Fair enough, I didn't explain this very thoroughly in the comment in the code.

My app has a Blaze template that embeds a React component using react-template-helper.  The React component manually manages the lifecycle of a subscription: it's created in `componentDidMount`, stopped and recreated in `componentWillReceiveProps` if the relevant props have changed, and I see now I'm missing the stop in `componentWillUnmount`.  However, the subscription gets associated with the computation created by react-template-helper so that it gets automatically stopped when the computation is invalidated, which I don't want.

A sub-component of the React component reads a ReactiveVar during rendering.  When this ReactiveVar changes, the react-template-helper computation gets invalidated and the subscription is stopped.  react-template-helper calls ReactDOM.render again, but the main React component does not create another subscription because it assumes the previous one still exists.  The data from the stopped subscription gets removed from the local collections, causing the app to break.

I see now that the value read from the ReactiveVar is not actually being used, so if I removed the read, the problem would probably go away, but that code was written by someone else and I wanted to wait and let them clean it up.  In the meantime, I think this mistake should not cause the subscription managed by the main component to get unexpectedly stopped.  Also, I think the same problem would occur if the Blaze component's data changed and as a consequence, one of the props of the main React component that does not affect the subscription changed, though I believe this currently doesn't happen in my app.

mattmccutchen:
I discovered another code path that causes a similar problem: `ReactMeteorData` calls `component.forceUpdate()` from within a Tracker computation.  I updated the pull request to address that one too.

filipenevola:

PR #260, opened 2018-09-01
Add displayName

merlinpatt:
Having a wrapped display name will make it easier to debug with React DevTools. Currently every wrapped component shows up as ReactMeteorDataComponent, which isn't very helpful. Now it will show up as `withTracker(SomeComponent)`


StorytellerCZ:
LGMT!

filipenevola:

PR #268, opened 2019-02-24
Remove react-meteor-data's dependency on  check-npm-dependencies

moberegger:
Removes the dependency on [`tmeasday:check-npm-versions`](https://atmospherejs.com/tmeasday/check-npm-versions). This will save ~10kb in the bundle payload and save a small runtime cost.

An argument can be made that `check-npm-modules` doesn't provide much value when compared to what the Meteor build tool already does. Without this, Meteor will already display an error message
> Cannot find module 'react'

In my opinion, that is clear enough and the error provided by check-npm-versions doesn't provide much more insight - at least not enough to warrant the weight it adds to the production build. Furthermore, if you are using this package, it is very likely that you already have `react` installed; I can't imagine a developer getting very far without having `react` installed... at least not far enough to use `withTracker` in any capacity. 

Whilst it is nice to get this error at build time during development, I don't think the tradeoff is worth it, especially considering that it would only be helpful in very rare circumstances.

### DIFF
--- a/packages/react-meteor-data/createContainer.jsx
+++ b/packages/react-meteor-data/createContainer.jsx
@@ -3,7 +3,6 @@
  */
 
 import { Meteor } from 'meteor/meteor';
-import React from 'react';
 import connect from './ReactMeteorData.jsx';
 
 let hasDisplayedWarning = false;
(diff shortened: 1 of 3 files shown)


yched:
#262 will require React 16.8+ (the PR currently bumps the version in the current  checkNpmVersions() call)

Do we have other ways to express that if we remove check-npm-modules ?

filipenevola:
closing because of react-meteor-data@2.0.0 https://github.com/meteor/react-packages/pull/273

Issue #267, opened 2019-02-12
Turning react-meteor-data to a React hook


fawzisf:
i thinks it is possible to ditch the HOC in favor of the new hooks, https://reactjs.org/docs/hooks-overview.html#-building-your-own-hooks
this would make the code much easier to maintain and share.
what do you think ?

fawzisf:
#263 looks really promising in this direction.

dburles:
Absolutely. It’s being tracked at #262

filipenevola:
closing because of react-meteor-data@2.0.0 https://github.com/meteor/react-packages/pull/273

PR #261, opened 2018-10-26
[WIP] Update lifecycle methods for newer React

StorytellerCZ:
Starting on a fix for #252 and #256 (and maybe #242)
Sadly it isn't as easy as replacing it with the recommended new methods (in short that breaks everything).
Moving `componentWillMount` functionality to `constructor` fixes that part, but `componentWillUpdate` has no easy backward compatible solution. So at least I'm presenting here this half fix and hopefully I can figure out something later or someone else can finish what I started. :thinking:

Issue #256, opened 2018-05-24
 Unsafe lifecycle methods were found within a strict-mode tree


nicubarbaros:
componentWillMount: Please update the following components to use componentDidMount instead: ReactMeteorDataComponent

componentWillUpdate: Please update the following components to use componentDidUpdate instead: ReactMeteorDataComponent

Following errors with react 16.3 and above

StorytellerCZ:
Now these warning are visible outside of strict mode as well due to upcoming React 17.

filipenevola:
closing because of react-meteor-data@2.0.0 https://github.com/meteor/react-packages/pull/273

Issue #252, opened 2018-03-30
React 16.3 => migration from componentWillMount and componentWillUpdate


fabien-h:
Not blocking at the moment ; but we have some ugly warning :)

softwarecreations:
Reproduce deprecation warnings with 3 commands:
```
curl https://install.meteor.com/ | sh
cd /tmp; meteor create --react foo
cd foo; meteor
```
Warnings produced by react 16 in the automatically generated `package.json`
```
    "react": "^16.9.0",
    "react-dom": "^16.9.0",
```
![image](https://user-images.githubusercontent.com/28704571/62879109-b1285f80-bd2a-11e9-9cd2-075adf48b875.png)

Finding offending code:
```
grep -rn ~/.meteor -e "componentWill\(Mount\|Update\)"

~/.meteor/packages/react-meteor-data/.0.2.16.w4vu92.v54d++os+web.browser+web.cordova/os/ReactMeteorData.jsx:127:  componentWillMount() {
~/.meteor/packages/react-meteor-data/.0.2.16.w4vu92.v54d++os+web.browser+web.cordova/os/ReactMeteorData.jsx:134:  componentWillUpdate(nextProps, nextState) {
```

https://github.com/meteor/react-packages/blob/b55f96d0c668bc8379f913011a85647e2907667b/packages/react-meteor-data/ReactMeteorData.jsx#L127

Twisterking:
Will there be any updates on this? Or do I have to fork it and rename it to `UNSAFE_...` myself?

dburles:
https://github.com/meteor/react-packages/pull/273

filipenevola:
closing because of react-meteor-data@2.0.0 https://github.com/meteor/react-packages/pull/273

PR #271, opened 2019-07-18
Create useTracker hook and reimplement withTracker HOC on top of it 

CaptainN:
This PR is a roundup of the discussion in #262, and supersedes: #242, #252, #256, #263, #266, #267, #268.

It does the following:
- Create a base hook, `useTracker` to for use as an alternative to `withTracker` in hooks based projects.
- Rewrites `withTracker` on top of `useTracker`, which sures up `withTracker`'s compatibility with React strict mode and the forthcoming Suspense and concurrent mode (issues #256, #252, #242 / PR #261)
- Updates docs!
- Removes deprecated `createContainer`.
- Above #262, this PR retains the synchronous rendering behavior in `withTracker` and in `useTracker` if no `deps` are specified, but also allows asynchronous reactivity after `firstRun` if `deps` are supplied, and properly responds to `deps` changes.

There were not any unit tests, but I may write some if I have time. I'd especially like to test some of the lifecycle stuff, and make sure reactivity persists. I don't think that should hold up a beta release.

Speaking of release - since this bumps the react version requirement to 16.8, there has been some discussion about how to properly release this with a major version bump. One suggestion was to bump the old version to 1.0.0 which can receive updates separately, then release this update as 2.0.0.

CC @hwillson @menelike @yched


menelike:
@CaptainN 

Maybe it makes sense to wait for a final solution to https://github.com/risetechnologies/react-packages/pull/2

IMHO this need to control the computation from both the computation and from the react component is absolutely valid. I definitely expect an API change when this will be introduced, so we should add that as well to avoid breaking changes.

I'll propose my solution this weekend.

CaptainN:
@menelike My favored solution to that would be additive, and would not require us to forestall the release of this to beta, or muddy up this issue, unless you think it's vital to absolutely change the API significantly. I'll add my proposal to the other issue thread.

CaptainN:
I generally don't think we should hold up beta on this, or that we should significantly change the API, for a couple of reasons:

- The API right now for this hook is clean, and I like that.
- In the old `withTracker` no one ever had access to the computation and it was fine.
- If someone does want access to the computation, they are likely jumping outside the react lifecycle anyway, and doing things imperatively. I don't know if we need to do more than simply provide them a way to get access to whatever the current instance is, and clean-up before it changes. My suggestion in risetechnologies#4 takes care of that. Additional, more robust tools can even be added on top of that.

Of course, as always, I can be convinced.

packages/react-meteor-data/useTracker.js
```diff
@@ -121,7 +119,7 @@ function useTracker(reactiveFn, deps) {
             runReactiveFn();
           }
           // use a uniqueCounter to trigger a state change to force a re-render
-          forceUpdate(++uniqueCounter);
+          forceUpdate(counter + 1);
```
menelike: Nice catch! (damn, why haven't I thought of this)

CaptainN:
In doing a final review myself, I noticed some things in the readme that are not quite accurate anymore (`deps` is optional now - recommended, not required). I'll rewrite that over the weekend. Additionally, there is one more feature @menelike and I have been trying to reconcile (allowing access to the computation). That will also hopefully be settled over the weekend. Heck, maybe I'll even sneak in some unit tests. It's Friday anyway - let's review Monday.

yched:
Hi there, sorry for the delay, I'm slowly catching up here.

I do get what the issue with my current code in #262 is (when the reactiveFn deps change, there is a transcient render with the new deps values but the old reactive result derived from the previous deps, before useEffect kicks in). I'm trying to figure out how I'd approach it while also following what you guys discussed and writing down some points of reply (waiting to see if they are still relevant for your final approach here)  

Getting there, please bear with me :-)

yched:
Still reading up, but I see a bug with the PR here, where useTracker seems to loses the reactivity when a deps array is provided :

```js
import React from 'react';
import { ReactiveVar } from 'meteor/reactive-var';

// Reactive func toggles between 0 and 1 every second:
const myVar = new ReactiveVar(0);
setInterval(() => myVar.set(1 - myVar.get()), 1000);
const reactiveFn = () => {
  const value = myVar.get();
  console.log(`reactiveFn - value ${value}`);
  return value;
};

// Component just outputs the reactive value:
function TestReactivity() {
  const value = useTracker(reactiveFn, []);
  console.log(`render - value ${value}`);
  return <div>Value: {value}</div>;
}
```

When mounting this, you see the component displays 0, then 1, and stays at 1.
In the console:
```
reactiveFn - value 0
render - value 0
reactiveFn - value 1
render - value 1
reactiveFn - value 0
render - value 0       (this render somewhat never makes it to the DOM ?) 
reactiveFn - value 1   (the subsequent reactive updates trigger no render)
reactiveFn - value 0
reactiveFn - value 1
(...)  
```

It works normally if the useTracker call omits the deps param (`const value = useTracker(reactiveFn);`

CaptainN:
I'll try and turn that into a Tinytest.

CaptainN:
Your test component is working correctly in my quick limited testing. Are you sure you have the latest version from the hooks branch? 

image

CaptainN:
@yched The problem was that the counter var was enclosed, and thus the computation handler couldn't see the updates to it, so forceUpdate was always setting the same value (0 + 1) - resulting in no forced update. I put it in a reference, and that fixed the problem. An older version used a package level variable, but I figured we can just use the counter we already have from the useState hook - forgetting the reference issues.

Anyway, it's fixed, please give it another go. My TinyTests didn't detect that - I'll make sure to add something that would detect that in case I ever break it again.

menelike:
@CaptainN too be honest, the new counter is more unintuitive then before (and now I remember why I did not use the state in the first place). Was your change really worth it taking maintainability into account? It's really hard to know why it needs to be that way.

I guarantee that a future refactoring thin `what the heck, I'm throwing away the ref counter and gonna use the state which is already there`. Fun fact: This actually happened when you used the state and it was me not writing a proper comment on why the counter is in the global state. So shame on me ;)

CaptainN:
@menelike I almost put your solution back, but I preferred this approach for two reasons - one is that it doesn't store a value in the package space, which I generally prefer to avoid, and the other is that although - EXCEEDINGLY unlikely to happen, there is a possibility that the counter can increment far enough to come back around, and cause a false match in some hook instance somewhere. (if there is a rapidly updating component that burns through all the available 32bit integers, and there are other hooks in the app, there can be a misfire.) I know the possibility is remote, but I prefer solutions that are 100% guaranteed to work. Maybe that's my OCD!

menelike:
@CaptainN That is fine, I can understand your point of view. Just make sure that whoever is gonna work on that code won't fall into the same pit. A proper code comment should cover that well.

menelike:
@CaptainN One final thought, doesn't it make sense to just use `ref.counter` and skip `refs.counter = counter`? So in the end, just moving the counter for the global state to instance state without touching any logic.

packages/react-meteor-data/useTracker.js
```diff
@@ -0,0 +1,155 @@
...
+// 34ce57ae751e0952fd12ab532a3e5694445897ea/packages/
+// react-reconciler/src/ReactFiberHooks.js#L307-L354
+// used to replicate dep change behavior and stay consistent
+// with React.useEffect()
+function areHookInputsEqual(nextDeps, prevDeps) {
+  if (prevDeps === null || prevDeps === undefined || !Array.isArray(prevDeps)) {
+    return false;
+  }
+
+  if (!Array.isArray(nextDeps)) {
```
CaptainN: Should this also check if `nextDeps` is falsy? If `nextDeps` is passed in as `null`, `undefined`, or `false` this sill trigger a positive match, but I think that's not correct at least for the warning (the return value seem correct).

CaptainN:
@menelike The `useState` method makes it easy to set the default - I guess we could do something like:

```js
  const { current: refs } = useRef({ counter: 0 });
  const [, forceUpdate] = useState(0);
```

Thinking low level, assigning the result of `useState` which is always allocated, even if we don't assign it to a local constant, to a property on refs, is probably slightly more efficient than instantiating that object with a counter allocation on every pass (`{ counter: 0 }`) - but it's really six of one, half a dozen of the other.

I had also thought of just incrementing the enclosed value (use let/var with `useState(0)`, then `forceUpdate(++counter)` inside the callback), but that felt hacky, and difficult to read.

menelike:
@CaptainN Now I understand, in that case, this is absolutely valid!

yched:
Yup, works now :-)

No need to bother with a maintaining a counter though, forceUpdate can be done with just:
```js
const [, forceUpdate] = useState(0);
forceUpdate(x => x + 1)
```

(some even externalize it as `const useForceUpdate = () => { const [, forceUpdate] = useReducer(x => x + 1, 0); return forceUpdate}` to make the consuming code clearer - see https://reactjs.org/docs/hooks-faq.html#is-there-something-like-forceupdate)

CaptainN:
I added the `useReducer` method described in the hooks-faq. At this point we are really optimizing around the edges. Even the two review questions I left for @menelike (@yched if you have any insights that would be welcomed!) are pretty minor. (It's basically asking if deps behavior should always be "off" when any falsy value is set, or only when "undefined" or "null" is set.)

Other than that, I think this is ready for review.

yched:
Well, getting to the meat of it, my concern with the approach in this PR is that it does run side effects directly in render (setting the Tracker computation, which itself can include setting DDP subscriptions - and reversedly stopping the computation and attached subscriptions). React docs explicitly state side effects in render is a big no-no :-/.

The way I understood it, with the forthcoming "time slicing" feauture, React can start rendering a tree, then discard it to instead render some higher priority update in another part of the UI first, then start again rendering the 1st tree from scratch. This only works if renders are "pure" and can be ditched and redone as react's scheduler sees fit: if renders that get discarded did set up computations and DDP subscriptions, those would stay dangling.
In short : React doesn't guarantee anymore that all renders are followed by an actual mount and useEffect cycle. That's the whole point of the deprecation of the willMount / willUpdate, and the reason why useEffect() runs at the didMount / didUpdate : you don't know when it *will*, you just know when it *did*.

"No side effects in render" were the reason for the monkey-patching that seemed to puzzle you in https://github.com/meteor/react-packages/pull/262#issuecomment-504128244. Now, as you guys pointed out, #262 does have annoying shotcomings, and I'm trying to think of a way to solve them, but I don't think this approach here flies either ?

yched:
Today's blog post from Dan Abramov happens to [mention that "render idempotency" assumption](https://overreacted.io/algebraic-effects-for-the-rest-of-us/#how-is-all-of-this-relevant-to-react), and it reminded me it's actually not just for the "time slicing" feature : React Suspense itself is based on "abandon rendering of a tree if a component throws a Promise, start rendering that tree again when the Promise resolves".

minhna:
does it support server rendering? current withTracker works with server rendering very well.

packages/react-meteor-data/useTracker.js
```diff
@@ -0,0 +1,155 @@
...
+  );
+}
+
+// inspired by https://github.com/facebook/react/blob/
+// 34ce57ae751e0952fd12ab532a3e5694445897ea/packages/
+// react-reconciler/src/ReactFiberHooks.js#L307-L354
+// used to replicate dep change behavior and stay consistent
+// with React.useEffect()
+function areHookInputsEqual(nextDeps, prevDeps) {
+  if (prevDeps === null || prevDeps === undefined || !Array.isArray(prevDeps)) {
```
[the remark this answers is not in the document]
menelike: I initially did this because of https://github.com/facebook/react/blob/34ce57ae751e0952fd12ab532a3e5694445897ea/packages/react-reconciler/src/ReactFiberHooks.js#L318 and https://github.com/facebook/react/blob/34ce57ae751e0952fd12ab532a3e5694445897ea/packages/react-reconciler/src/ReactFiberHooks.js#L878

Which only checks for `null` and `undefined`. Afaik `if (!prevDeps || !Array.isArray(prevDeps) {` should work as well since `0` is not expected anyway, also might be easier to think, I agree with that change.

packages/react-meteor-data/useTracker.js
menelike: This was taken from https://github.com/meteor/react-packages/blob/bff72c7c291743acd5348f81846ea2679000666c/packages/react-meteor-data/useTracker.js#L49

```JS
  if (!Array.isArray(nextDeps)) {
    if (Meteor.isDevelopment) {
      warn(
        'Warning: useTracker expected an dependency value of '
        + `type array but got type of ${typeof nextDeps} instead.`
      );
    }
    return false;
  }
```

I am not sure if I understand you correctly here, this should take care of all scenarios?:

```JS
!Array.isArray(undefined) => true
!Array.isArray(null) => true
!Array.isArray(0) => true
!Array.isArray([]) => false
```

menelike:
@yched 

> but I don't think this approach here flies either ?

IMHO this PR fulfills the *current* state of React and keeps backward compatibility, but I agree, it's not a sustainable solution.

WDYT about this https://github.com/risetechnologies/react-packages/commit/0001f4903306633d34a8cc021b8dfe0813a4beba

The primary goal was to workaround

> if renders that get discarded did set up computations and DDP subscriptions, those would stay dangling.

```JS
function useTracker(reactiveFn, deps) {
  const { current: refs } = useRef({});

  const [, forceUpdate] = useReducer(fur, 0);

  const dispose = () => {
    if (refs.computation) {
      refs.computation.stop();
      refs.computation = null;
    }
  };

  if (!refs.didMount) {
    // No side-effects are allowed when computing the initial value.
    // To get the initial return value for the 1st render on mount,
    // we run reactiveFn without autorun or subscriptions.
    // Note: maybe when React Suspense is officially available we could
    // throw a Promise instead to skip the 1st render altogether ?
    const realSubscribe = Meteor.subscribe;
    Meteor.subscribe = () => ({ stop: () => {}, ready: () => false });
    Tracker.nonreactive(() => {
      const data = reactiveFn();
      if (Meteor.isDevelopment) checkCursor(data);
      refs.trackerData = data;
    });
    Meteor.subscribe = realSubscribe;
  } else if (!areHookInputsEqual(deps, refs.previousDeps)) {
    // this is called like at componentWillMount and componentWillUpdate equally
    // in order to support render calls with synchronous data from the reactive computation
    // if prevDeps or deps are not set areHookInputsEqual always returns false
    // and the reactive functions is always called
    // if we are re-creating the computation, we need to stop the old one.
    dispose();

    // store the deps for comparison on next render
    refs.previousDeps = deps;

    // Use Tracker.nonreactive in case we are inside a Tracker Computation.
    // This can happen if someone calls `ReactDOM.render` inside a Computation.
    // In that case, we want to opt out of the normal behavior of nested
    // Computations, where if the outer one is invalidated or stopped,
    // it stops the inner one.
    refs.computation = Tracker.nonreactive(() => (
      Tracker.autorun((c) => {
        const runReactiveFn = () => {
          const data = reactiveFn();
          if (Meteor.isDevelopment) checkCursor(data);
          refs.trackerData = data;
        };

        if (c.firstRun) {
          // This will capture data synchronously on first run (and after deps change).
          // Additional cycles will follow the normal computation behavior.
          runReactiveFn();
        } else {
          // If deps are falsy, stop computation and let next render handle reactiveFn.
          if (!refs.previousDeps) {
            dispose();
          } else {
            runReactiveFn();
          }
          forceUpdate();
        }
      })
    ));
  }

  // stop the computation on unmount only
  useEffect(() => {
    refs.didMount = true;
    if (Meteor.isDevelopment
      && deps !== null && deps !== undefined
      && !Array.isArray(deps)) {
      warn(
        'Warning: useTracker expected an initial dependency value of '
        + `type array but got type of ${typeof deps} instead.`
      );
    }

    return dispose;
  }, []);

  return refs.trackerData;
}
```

*Note: This is untested and drafted code, the monkey patching part needs to be solved in Meteor or MeteorTracker later on*

CaptainN:
@yched The more I learn about the direction of hooks, the more I think I should just use vue...

Since we can't have a practical pre-render hook because academic purity is more important (this reminds me SO much of the peak of AS3 development, right before it all collapsed), we could just accept double rendering with `useLayoutEffect`. This hook will fire after render, presumably after React has decided whether it'll keep the current output or not, then re-render the tree again sychronously, because that's efficient. It'll all work synchronously though, and the full life cycle will be retained, so the end result should work for us. What do you think?

CaptainN:
@menelike It's probably easier to use `useLayoutEffect` but I think going back to @yched's early approach would work too. In that solution, care was taken to avoid creating assets which couldn't be canceled (subscriptions) during render, then just re-running everything in the effect hook. We could work around double render issue by not forcing render on `firstRun`. In other words, a hybrid approach - store the data in a ref like our new fork and use forceUpdate for actual updating. I'll work up a patch.

CaptainN:
I pushed an update which solves the problem using the "hybrid" model - it runs a side-effect free version of the computation (@yched's earlier solution) and then sets everything else up in `useEffect`. This causes the user's `reactiveFn` to run at least 2 times on mount, possibly more with Suspense.

This also breaks my TinyTests (which I could use help with to integrate properly into the main tests - which don't seem to run correctly, at least on my Windows machine).

I may prefer the `useLayoutEffect` method, which would cause a double react render, but would only invoke the user's `reactiveFn` once, would help protect the user from doing side effects in their own code (actually this is a big reason for `useLayoutEffect`) and would avoid monkey patching the core subscribe methods.

menelike:
This issue is really a pain in the a**.

Please don't get me wrong, all I can say is that until React decides to finalize their API I'll stick to what we had before since it's *currently* not an issue. In addition, the performance was really bad with the last approach, and after all, what's the point of hooks then anyway.
One of the most important reasons in this context is a better performance than before. It doesn't look like better or at least the same performance can be achieved with how react Fibers will work.

CaptainN:
Okay, `useLayoutEffect` kind of works, but we'd have to accept 1 render pass with no data, which is burdensome (the user would have to always check for `null` or something) - the react team may not care about burdening their users, but I do. We would probably also need new logic to deal with deps changes, especially when no deps are supplied. I think the solution we have, where we run it synchronously with mitigated side-effects, then set everything else up in useEffect is better, even if it means the `reactiveFn` has to be re-run in the name of functional purity.

@menelike Please try the last version, see if it solves your issue and if it has acceptable performance. I think it will.

If anyone is curious, here is the `useLayoutEffect` version (and it shows how nice the code would be if there was something like a `useSyncEffect` or something, grumbles...):

```js
/* global Meteor, Package, Tracker */
import React, { useLayoutEffect, useRef, useState } from 'react';

// Use React.warn() if available (should ship in React 16.9).
const warn = React.warn || console.warn.bind(console);

// Warns if data is a Mongo.Cursor or a POJO containing a Mongo.Cursor.
function checkCursor(data) {
  let shouldWarn = false;
  if (Package.mongo && Package.mongo.Mongo && data && typeof data === 'object') {
    if (data instanceof Package.mongo.Mongo.Cursor) {
      shouldWarn = true;
    } else if (Object.getPrototypeOf(data) === Object.prototype) {
      Object.keys(data).forEach((key) => {
        if (data[key] instanceof Package.mongo.Mongo.Cursor) {
          shouldWarn = true;
        }
      });
    }
  }
  if (shouldWarn) {
    warn(
      'Warning: your reactive function is returning a Mongo cursor. '
      + 'This value will not be reactive. You probably want to call '
      + '`.fetch()` on the cursor before returning it.'
    );
  }
}

// Used to create a forceUpdate from useReducer. Forces update by
// incrementing a number whenever the dispatch method is invoked.
const fur = x => x + 1;

function useTracker(reactiveFn, deps = []) {
  const computation = useRef(null);

  const [trackerData, setTrackerData] = useState(null);

  const dispose = () => {
    if (computation.current) {
      computation.current.stop();
      computation.current = null;
    }
  };

  // stop the computation on unmount only
  useLayoutEffect(() => {
    if (Meteor.isDevelopment
      && deps !== null && deps !== undefined
      && !Array.isArray(deps)) {
      warn(
        'Warning: useTracker expected an initial dependency value of '
        + `type array but got type of ${typeof deps} instead.`
      );
    }

    // Use Tracker.nonreactive in case we are inside a Tracker Computation.
    // This can happen if someone calls `ReactDOM.render` inside a Computation.
    // In that case, we want to opt out of the normal behavior of nested
    // Computations, where if the outer one is invalidated or stopped,
    // it stops the inner one.
    computation.current = Tracker.nonreactive(() => (
      Tracker.autorun((c) => {
        const data = reactiveFn();
        if (Meteor.isDevelopment) checkCursor(data);
        setTrackerData(data);
      })
    ));

    return dispose;
  }, deps);

  return trackerData;
}

// When rendering on the server, we don't want to use the Tracker.
// We only do the first rendering on the server so we can get the data right away
function useTrackerServer(reactiveFn) {
  return reactiveFn();
}

export default (Meteor.isServer ? useTrackerServer : useTracker);
```

CaptainN:
That github outage gave me a second to rethink some core assumptions (I hate to do this so close to what we hoped is the end). The problem is that we are trying to violate the react rule of not creating side effects during rendering, or at least to work around them. Setting up a computation always produces a side-effect. Basically, the entire Meteor data access model relies on setting up side effects.

Is it necessary then for the Meteor tool to try to run this synchronously in every case? I wonder whether we should at least consider not working so hard to provide 100% backward compatible behavior - or to make it opt-in, and encourage the use of a purer `useTrackerEffect` hook instead.

We could even implement it both ways - perhaps `useTracker` is an effect hook (no initial synchronous behavior, no monkey patching, safe for side-effects in reactiveFn) while `useTrackerCompat` has the synchronous behavior (and is used for `withTracker`), but must come with warnings about side-effects.

What do you think?

(I still think there should be a before-render-effects hook.)

menelike:
@CaptainN 

> @menelike Please try the last version, see if it solves your issue and if it has acceptable performance. I think it will.

Are you sure that you'll fetch the data on the first render? Haven't tested it, but `useLayoutEffect` isn't synchronous as well:  `If you’re migrating code from a class component, note useLayoutEffect fires in the same phase as componentDidMount and componentDidUpdate.
` 

I think you're right. There is not a backward-compatible solution which will work nicely with how react will handle render phases in the future, this topic might be needed to view from a different angle.

```
This builds on a React feature called “Suspense”, which is in active development for the data fetching use case. The interesting part, of course, is that the data might not yet be in the movieCache — in which case we need to do something because we can’t proceed below. Technically, in that case the read() call throws a Promise (yes, throws a Promise — let that sink in). This “suspends” the execution. React catches that Promise, and remembers to retry rendering the component tree after the thrown Promise resolves.
```

from https://overreacted.io/algebraic-effects-for-the-rest-of-us/#how-is-all-of-this-relevant-to-react might give us some tools in the future to solve this completely different.

But still, is it worth to already deal with circumstances which are just theory right now? After all, this decision has to be made my MDG, for the time being I'll use with what has been achieved before, it performs very well, helps to structure code beautifully and is even easy to deal with in conjunction with flow type. In comparison to `withTracker` clearly a massive benefit.

Also, since we had no breaking change this could have been a perfect upgrade of `withTracker`, which btw. is also using deprecated methods anyway. IMHO this PR should target only `hooks` and nothing else.

CaptainN:
@menelike The latest version in the PR doesn't use `useLayoutEffect` - it's the mentioned hybrid approach. It runs synchronously with the side-effect mitigation that @yched had in the initial PR of all this, then it sets up all the rest in `useEffect` later.

It avoids updating the state on first-render so we don't get a double render - this assumes the data won't change between the two - is that safe? If it's not safe, I say we go back to the previous non-Suspense behavior (as @menelike suggested), and wait for Suspense to be closer to finalized before we try this again.

menelike:
@CaptainN 

I am currently examining your recent changes, I am not sure if I miss something, but it looks like on every dep change you call the reactive function two times?

On dep change both https://github.com/meteor/react-packages/blob/b9d69114b052a3a58a78f33479762e5e5be90864/packages/react-meteor-data/useTracker.js#L100
and https://github.com/meteor/react-packages/blob/b9d69114b052a3a58a78f33479762e5e5be90864/packages/react-meteor-data/useTracker.js#L153 will result in a call of `runReactiveFn` ?

CaptainN:
@menelike Yes, it does call the `reactiveFn` two times for every deps change. I guess for the case where we have no deps this is actually a problem. I actually think the race condition this sets up (because I'm ignoring the change on first run from the `useEffect` hook, the second run of `reactiveFn`) is problematic too. For both of these reasons I think I should revert this back to the previous synchronous behavior, even though we are relatively sure it won't work correctly with Suspense. I have to assume that when React ships with Suspense, they'll have some kind of solution to the problems we've identified there, and we can revisit at that time.

packages/react-meteor-data/useTracker.js
CaptainN: Oh, you are right, this is fine.

menelike:
After all, all we need to know when supsense finally comes:

1. Has the render been thrown away?
2. Then either use the cached result (if no deps have changed), tear it completely down if an unmount happened in the meantime or reinitiate the computation when deps have changed in the meantime.

All we need is just one parameter ;) 

Thanks for all your efforts and always trying to make the best out of it @CaptainN , I highly appreciate it!

packages/react-meteor-data/useTracker.js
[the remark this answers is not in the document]
CaptainN: I was thinking that a "falsy" value like null or undefined is actually valid. This check though will kick up a warning if null or undefined are used.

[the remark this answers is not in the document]
CaptainN: Oh I see that React doesn't allow the value to change. This is a mistake IMHO, because React will sometimes reuse components in ways that are not always clear (in your react-router `Switch` case for example). I think we should assume those values can change at any time (and our component does a great job of allowing that - with the possible exception of the `useEffect` hooks, which takes unfiltered `deps` - more edge case shenanigans from upstream...)

yched:
> The more I learn about the direction of hooks, the more I think I should just use vue...

Well, to be fair I don't think the tricky thing is about hooks per se, but rather the deprecation of the willMount / willupdate lifecyles - see the previous unsuccessful attempts at moving the withTracker HOC away from them since their deprecation was announced, before hooks were even on the radar. It's just tricky, and I think we would have the same difficulties making it work with a class-based version.

(FWIW, the React team deprecated those lifecycles because they're betting that time slicing and suspense will make apps feel more "reactive", even at the cost of possibly more renders. In the "willUpdate -> didUpdate" discussions and FAQs, they repeatedly downplayed the impact of extra renders; Whether that's a winning bet remains to be seen, but I'd tend to agree that the ability to prioritize renders seems to be worth a couple extra renders).

yched:
Also : curious why you think (thought ?) that useLayoutEffect might be a better option ?

CaptainN:
@yched I hadn't thought through `useLayoutEffect` thoroughly enough - I figured, it would get me synchronous rendering, which it does, but it doesn't solve the double render invocation problem at all. Was fun to play with though!

I agree, the problem is they got rid of any reliable way to do stuff which may set up side-effects before rendering. The way it is now is actually fine - why bother with pre-render hooks, when you can just do stuff in the render method before you return the rendering code. But if they make that unreliable too, then there really is no way to set this stuff up without waiting for additional renders. Now that may not really be a problem (I'm sure the engineers on React would argue the concurrency gains and efficiency gains are worth the trade-off), but it does mean we have to swallow a second render in our specific case (and we can't have a perfectly backward compatible way to do this for `withTracker`).

CaptainN:
@yched I actually like what they are doing with React, but I really think they are being overly dogmatic by not providing a way to set up effects until after the first render. I don't believe it's necessary to limit us in that way - I think they could provide a hook that is convenient, and still make it efficient. I get the sense that many of their decisions are based on a pair of higher level concerns - 1) functional purity (in AS3 it was OOP/GOF purity), and 2) trying to protect the developer users from themselves (in AS3 it was endless design patterns with inversion of control and all that crazy stuff - end users just ended up pulling out the logic of those frameworks in their local classes).

Especially #2 is an error in judgement - if you don't provide developers with straight forward ways to do what they want (in this case, a workable alternative to componentWillMount), they'll make up their own terrible ways to do it, and that'll just lead to more problems later (I'm even specifically arguing for one such case in this very PR). This can already be seen with the `useMemo` - which would have been ideal for some of what we put together here - except it doesn't make a semantic guarantee (if if it currently does behave that way). So we ended up having to re-invent a few wheels, even pulling out internals the way @menelike did. That can't be what the React team intended, but if you google around for the problem, you'll find a ton of half-implemented alternatives to `useMemo` - and actually, you'll find just as many examples where people are just using `useMemo` as if it does have a semantic guarantee. These are like grenades with the pins pulled, just waiting to go off all over people's react projects.

Of course, it's entirely possible (maybe even probable) that by the time `componentWillMount` is really removed, and suspense is the new norm, the React team will have worked out some solution (throwing a promise or something). Until then, I think our current approach is defensible.

CaptainN:
Just one more note (sorry for the spam) - in our case it's less that I'm worried about re-renders. I'm more concerned about having a nice clean and easy to use API.

We could for example, return a null value until the `withEffect` hook runs to set up our computation and then document that behavior and expect users to always check for null. Then then hack around that for backward compatibility in `withTracker` - but really, this shouldn't be such a hard problem to solve. The Meteor reactive func does set up side-effects, and we do need to keep track of that, but it also runs right away and returns data that is often available right away. There's no reason in the world this pattern should be so hard to set up.

CaptainN:
@menelike I agree that it shouldn't be hard to solve, and Suspense may have a way to do that. The main problem is not that we are creating side-effects in render - it's that we need a way to clean up when the `useEffect` hook is not run. Maybe the promise it throws has an error state, or something we can use to detect when the current render is tossed, and then clean up our side-effects - that would solve it cleanly.

CaptainN:
@hwillson In short this PR is ready for review.

- It performs well with the current version of React (16.8.x), and solves the problems listed in the main PR text.
- It also has some tests - however, I'm not sure how to integrate those with the main tests in the root of the project, and I could use some help with that. (The problem may just be that I'm on Windows, and the tests don't work on Windows.)
- It does not solve future problems associated with Suspense/time-slicing/whatever, but I think it's defensible to punt on that until Suspense is closer to finalized.

yched:
You guys move too fast for me :-D

FWIW, the "mixed" implementation in b9d6911, which was reverted since then, was more or less what I had in mind: whenever we have to setup a new computation (1st render or deps change), we run a side-effect-free version of the reactiveFn to get the sync return value for the current render, and setup the actual computation / subscription in useEffect. I had a couple remarks / simplifications on the code itself, but that seemed to me pretty much the best we can do at the moment.

I'm not really worried about the extra renders caused by useEffect() **if** they are triggered with setState(), since, unlike forceUpdate(), setState(sameValue) (which is the case most of the time with the 2nd render from useEffect) will re-render the compoent but not its children. I think a version of b9d6911 can be written that way, I'll try to do that (maybe in #262 ?), and then we can compare performance ?

Last thing for now : whatever the approach ("impure"/HEAD or "mixed"/b9d6911) : is there really no way to base our deps comparison on useMemo() rather than duplicating its low-level logic ? I know @CaptainN pointed from [the docs]((https://reactjs.org/docs/hooks-reference.html#usememo) "[use it] as a performance optimization, not as a semantic guarantee, React may choose to “forget” some previously memoized values and recalculate them on next render"), but is that really a deal breaker here ?

CaptainN:
@yched I would not be apposed to a hybrid solution based on `useMemo`. I prefer the current version mostly because it is almost perfectly optimized given current API constraints. But a `useMemo` approach even makes sense here, given how easy it is to rebuild the state. I think I'd still rather use the current version - having 2 initial renders for every `useTracker` (even if one is shallow) is to me not the best default, especially if there is a possible workaround eventually even with Suspense. If we have to, I'll accept it, but it's not my preference.

CaptainN:
I couldn't get the old tests to work as they were set up - and I noticed they were out of date anyway (they only tested the old `ReactMeteorData` mixin). So I ported them to my in-package testing setup. One set for `useTracker` and one set for `withTracker`, with some extra stuff for testing with deps specified. I also learned how to get some of my hook specific tests which rely on tracker reactive changes to stop stalling. Pretty spiffy.

Once again, @hwillson, I think this is ready for review! (Should I make a new PR again? heh)

CaptainN:
@yched Because I obviously can't help myself, I created a branch with `useMemo` - the trade off is that we lose the local deps comparison functions, but we add an additional hook (useMemo), and are kind of abusing it. It's only used for its deps support in this branch, and I had to use a local counter ref to make sure it updates the way we intend when deps are falsy (when that's the case, it's pure overhead). But it's not terrible - [check it out](https://github.com/CaptainN/react-packages/blob/hooks-useMemo/packages/react-meteor-data/useTracker.js) if you are interested.

packages/react-meteor-data/useTracker.js
menelike: @CaptainN On second thought, I think react is doing that only for performance reasons. Maybe early exiting the check on null or undefined is faster than Array.isArray(), I don't mind if you change it.

menelike:
@CaptainN I actually like the idea of `useMemo`, you don't even need https://github.com/CaptainN/react-packages/blob/hooks-useMemo/packages/react-meteor-data/useTracker.js#L51-L53 since `If no array is provided, a new value will be computed on every render.` (maybe I did miss something, I've already got too many headaches because if this topic ;) )

But as @yched already mentioned, `You may rely on useMemo as a performance optimization, not as a semantic guarantee.` which means we have no control at all (it's not gonna be fun to debug this).
Even if I like the approach, having something which is rock-solid is more important here, otherwise, it feels like building something on a swamp.

Ideally react would expose their internal checking algorithm at some point like they did with [shallow-compare](https://reactjs.org/docs/shallow-compare.html).

On a side note, I can't find the comment but you mentioned that passing deps like `{foo: 'bar'}` will always resolve in reinitializing the computation. Since we are already checking the deps I was thinking of doing a 2 level deep shallow comparison to solve that, but clearly, this is something for another PR. 
I am considering to publish an own package anyway, working in a large project this piece of code has a **huge** impact on performance.

packages/react-meteor-data/useTracker.js
menelike: You might be interested in https://github.com/risetechnologies/react-packages/commit/e3bc3656586b812728ae57fe91222e48cd00ad66 `areDepsValid`, but be careful this is WIP

packages/react-meteor-data/package.js
```diff
@@ -6,13 +8,17 @@ Package.describe({
   git: 'https://github.com/meteor/react-packages',
 });
 
-Package.onUse(function (api) {
+Package.onUse((api) => {
```
menelike: I just read `Note that currently the package.js file itself is processed as Javascript/ES5. This cannot be changed.`
https://github.com/meteor/meteor/tree/master/packages/tinytest#structuring-the-test-file

Not sure what or if something will fail, but you might want to consider using ES5 here.

CaptainN:
@menelike If I said something like that, I don't know if it's correct. deps is always an array, so you wouldn't pass something like `{ foo: 'bar' }` to it directly. I suppose if you get an object back from some source, then put that in the array, it would always trigger re-compute, but I don't know if it's enough of a worry anyway. The optimization strategy for `useTracker` is identical to `useEffect`, `useMemo` etc. If relying on some object's `foo` property, you'd put that directly into the deps array: `[foo.bar, baz.bobo]`.

One of the nice things about this hook is when `deps` isn't set up correctly, the worst that will happen is it'll be no more inefficient than the old `withTracker` - that's pretty cool.

I like the shallow compare we have now for two reaso

packages/react-meteor-data/useTracker.js
CaptainN: I think it all matches now, does it not?

I would have thought a falsy check would be enough (`if (!deps && !Array.isArray(deps)) {`) - but the React code seems to specifically avoid checking `false`.

menelike: I am not really sure (working on this code every now and then is really confusing, always losing something in between). Right now I read: If the nextDeps is not an array, always warn. But nextDeps can actually be null or undefined per spec, so when in development it warns.

I think we might not need a change here because since prevDeps and nextDeps usually should always be the same type, e.g. once deps are null, they should stay null. Since the prevDeps check happens before and early returns false, this code *should* be able to stay like it is.

CaptainN: I actually think it should treat changes to deps as a fresh invokation every time (the react team gets this wrong in their code comment). Your `react-router` `Switch` example demonstrates why. Sometimes the React engine will re-use an instance in ways you don't expect - in those cases, it really is like starting over. I've kept that in mind all the way through the development of this, and I think it's lead to a rock solid code base. Here I think the issue is actually relatively minor - but maybe the fix is too:

```js
  if (!Array.isArray(nextDeps)) {
    if (Meteor.isDevelopment && deps !== undefined && deps !== null ) {
      warn(
        'Warning: useTracker expected an dependency value of '
        + `type array but got type of ${typeof nextDeps} instead.`
      );
    }
    return false;
  }
```

CaptainN:
Okay, one last wrinkle in this (trying to get to the end) - I created a version in a branch which contains a computation life-cycle function argument. It works like a `useEffect` method, where it runs whenever a new computation is created, then runs the function returned from that when the computation is destroyed.

```js
// This is extremely contrived!
function MyComponent ({ name }) {
  const reactiveDict = useRef()
  const value = useTracker(() => {
    // do reactive stuff
    return reactiveDict.current.get('key')
  }, [name], (c) => {
    // do something with the computation, or initialize stuff (in most cases there are better ways to do this)
    reactiveDict.current = new ReactiveDict(name, { key: 1 })
    const id = setInterval(() => reactiveDict.current.set('key', reactiveDict.current.get('key') + 1), 100)
    return () => {
      // the computation will be destroyed - now's our chance to clean up
      reactiveDict.current.destroy()
      clearInterval(id)
    }
  })
  return <div>{name}: {value}</div>
}

function TestMyComponent () {
  const [name, setName] = useState('n')
  useEffect(() => {
    const id = setTimeout(() => {
      setName(name === 'y' ? 'n' : 'y')
      return () => clearTimeout(id)
    }, 3000)
  }, [name])
  return <MyComponent name={name} />
}
```

`withTracker` never had access to the computation, so I don't know how necessary it is for real world use cases to get access to the actual computation. However, I can say that it's pretty useful to verify some computation cycle behavior in tests. I have also created a few hooks where it'd be nice to clean stuff up when the computation changes (or is disposed). I currently use `useEffect` for that - but if there's already a way to do it with `useTracker` then why not expose it?

Maybe I should create a PR from that branch (it is up to date with everything here) - what do you think? I only hesitate because I'd rather complete any conversation here, before creating yet another PR.

https://github.com/CaptainN/react-packages/blob/hooks-cleanup-2/packages/react-meteor-data/useTracker.js

@menelike @yched

packages/react-meteor-data/useTracker.js
menelike: looks perfect! 🚀

menelike:
@CaptainN I am too opinionated regarding how one should be able to control the computation.
I can think that this is a reasonable choice, I'd like to let others decide.

CaptainN:
@menelike I think you have a different problem (described in the other issue), where you want to control the computation. This new question is about whether it's worth it to expose its life cycle. I [commented on the other issue](https://github.com/risetechnologies/react-packages/pull/4#issuecomment-514691382) to continue that conversation there.

packages/react-meteor-data/useTracker.js
```diff
@@ -0,0 +1,155 @@
...
+
+        if (c.firstRun) {
+          // This will capture data synchronously on first run (and after deps change).
+          // Additional cycles will follow the normal computation behavior.
+          runReactiveFn();
+        } else {
+          // If deps are falsy, stop computation and let next render handle reactiveFn.
+          if (refs.previousDeps !== null && refs.previousDeps !== undefined
+            && !Array.isArray(refs.previousDeps)) {
+            dispose();
```
yched: I don't think I get that conditional : if previousDeps are neither null, undefined nor an array ? What would they be then ? Meaning, I don't get in which cases we go in that code branch currently ?
(and if the code is correct, then it's the comment that if off ?)

yched: Maybe it could be helpful to explain why we're looking at previousDeps rather than deps here ?

CaptainN: Valid types for deps are `null`, `undefined` or an array. Previously, I thought we were checking for all "falsy" values, but I think the react code only checks for the listed types (as does the above deps compare check). The comment is out of date.

I'm looking at `previousDeps` because the ref is at this point always the update to date version of the deps. This closure is always created right after deps change in response to that change. We could probably just check the enclosed `deps` instead of `previousDeps` - it doesn't really matter because they should always be the same inside this closure.

packages/react-meteor-data/withTracker.jsx
```diff
@@ -0,0 +1,16 @@
+import React, { forwardRef, memo } from 'react';
+import useTracker from './useTracker.js';
+
+export default function withTracker(options) {
+  return Component => {
+    const expandedOptions = typeof options === 'function' ? { getMeteorData: options } : options;
+    const { getMeteorData, pure = true, deps = null } = expandedOptions;
+
+    const WithTracker = forwardRef((props, ref) => {
+      const data = useTracker(() => getMeteorData(props) || {}, deps);
```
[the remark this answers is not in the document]
CaptainN: Good point, maybe I should just get rid of that.

packages/react-meteor-data/useTracker.js
```diff
@@ -0,0 +1,155 @@
...
+    ));
+  }
+
+  // stop the computation on unmount only
+  useEffect(() => {
+    if (Meteor.isDevelopment
+      && deps !== null && deps !== undefined
+      && !Array.isArray(deps)) {
+      warn(
+        'Warning: useTracker expected an initial dependency value of '
```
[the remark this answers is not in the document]
CaptainN: You are right, this is confusing. I'm not even sure it's correct. Verifying this is one of the reasons I asked about the computation life-cycle argument.

packages/react-meteor-data/useTracker.js
CaptainN: I fixed the logical error, and added some better comments

CaptainN: I fixed the logical error, and also took a harder look at it. It must be an optimization. The null and undefined checks must be early exits (as @menelike suggested) to avoid calling Array.isArray (maybe because that's often polyfilled), since Array.isArray would actually be enough on it's own, logically speaking.

packages/react-meteor-data/useTracker.js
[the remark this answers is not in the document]
CaptainN: I clarified some of this in recent commits/comments. I think it's even cleaner in hooks-cleanup-2 branch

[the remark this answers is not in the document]
yched: Would still be more legible if this was moved out of the way and into a Meteor.isDevelopment && checkDeps(deps) ?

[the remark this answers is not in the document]
CaptainN: Really all that's needed for most of these (some of them are slightly different) is `!Array.isArray(deps)` - checking for null and undefined is a performance optimization, which wouldn't be optimized if we used a function. For readability, just using `Array.isArray` is best - but I figured the optimization was worth it?

CaptainN:
Looking closer at React Suspense, it looks like it's mostly an async loading paradigm, which is not at all how Meteor works. To properly support that we'd probably want to do some more radical design changes to this hook (or creating another one). For example, a `useSubscription` hook could throw a promise, then when it comes back the following `useTracker` hook could run.

The problem comes if someone uses a Suspense feature along side our `useTracker` implementation - eg, they throw a promise after `useTracker`. I explained the problem fairly clearly in the [React ticket](https://github.com/facebook/react/issues/16175#issuecomment-515086584) on the subject.

I wonder if in the short term it's enough to explain that `useTracker` should come after Suspense hooks. The [example here](https://codesandbox.io/s/suspense-effect-no-cleanup-xwjjj), shows that the mock hook (`useInitialEffect`) works fine after the thrown promise.

The ideal solution is for us to somehow get a safe way to create or clean up the side-effect we might create before a promise is thrown. Otherwise we have to accept an inefficient default.

packages/react-meteor-data/useTracker.js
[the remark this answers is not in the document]
yched: sure, why not the optimization - but in itself it would deserve a comment because yes, anyone looking at this code is very likely to wonder "why not just Array.isArray() ?", and this is another argument for splitting that to a function, because we don't want to duplicate the comment as well :-D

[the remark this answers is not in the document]
CaptainN: I actually did change this particular check:

```js
  // stop the computation on unmount only
  useEffect(() => {
    // falsy deps is okay, but if deps is not falsy, it must be an array
    if (Meteor.isDevelopment && (deps && !Array.isArray(deps))) {
      warn(
        'Warning: useTracker expected an initial dependency value of '
        + `type array, null or undefined but got type of ${typeof deps} instead.`
      );
    }

```

CaptainN:
@yched Someone in the [react ticket](https://github.com/facebook/react/issues/16175#issuecomment-515110197) I opened suggested creating a custom Suspense component, to help us out with making our solution work with Suspense - I like the idea. What do you think?

We could do a custom `MeteorSuspense` or even something generic, and just put it on npm. He suggested two different models. (I'm still hoping React team will bundle something into react itself.)

CaptainN:
@yched This is really a worm in my brain. I had another idea - what if we just use a timeout between the synchronous portion of the hook, and the `useEffect` portion.

1. Synchronously, we start the computation as we do now, and start a timeout, with the cleanup method.
2. Then in `useEffect` we'd check to make sure to clear the timeout, and if the computation is still there, continue on - if not, rebuild it and rerender (like your original PR).
3. For reactive events, we could check the computation against the `refs.computation` to make sure that it's still valid (we could even dispose at that point).

Would that solve it? We would just need to set the timeout to some appropriate length - but it would probably not have to be very long. Maybe 50ms?

yched:
@CaptainN heh - that sounds a bit acrobatic, but right now I don't see where that would fail, actually... A timeout introduces another level of asynchronicity that comes with its own potential for interesting race conditions :-), but that definitely looks promising.

The penalty would be, if Suspense holds our component for more than (say) 50ms, then we'll stop the computation & subscriptions that were initiated, and will just start them again when it's actually mounted, that sounds acceptable. 

Not sure what would happen for reactive events triggered between the initial render and the actual mount, though. We wouldn't want them to trigger a rerender before useEffect runs and ensure our render was committed ?

CaptainN:
I'm out of time to play with it - but it might be possible to detect if the react component has been discarded or not from within the reactive function when the timeout happens - I'm not sure how to do that though.

yched:
Another thing that I've been thinking lately (speak af a brain worm...) : **maybe** having a version of **just** the withTracker HOC that is fully future-compatible with React Suspense - i.e. no componentWillXxx / no side effects in render / no double render - could be fairly simple : 

withTracker() is easier because a) we know the reactive func returns an object of props, and b) it wraps a component as a whole. So we could :
- have an internal useTrackerInternalForHOC hook (made up name of course) that just returns null on sync renders (no monkey-patching), and does Tracker.autorun() / setState() in use(Layout)Effect()
- have the wrapping component render nothing if the hook returns null, or the wrapped component with the returned props otherwise.

There's some handwaving here, I'd need to actually try it, but if it works it could give us some leeway to release an iso-API version of react-meteor-data to make existing apps ready for the depreations in React 16.8.7 and Suspense in 16.9 (which remains the most urgent goal IMO), and give us time to figure out the more generic useTracker() hook if it's not too clear yet how it would work with Suspense ?

CaptainN:
I actually got a version of this working - it's simpler than I thought, but there may be a race condition. It does this:

1. Set `isSuspended` to true by default in `useRef`
2. In `useEffect` set `refs.isSuspended` to false. If `useEffect` ran, the contract says its cleanup will run too.
3. In a timeout, see if `refs.isSuspended` is true - if so, dispose.

The main question I have (the race condition) is how long is the right amount of time to wait in setTimeout? Is there some guarantee of how long React will take before it executes `useEffect`? Another question is, is there another way to tell whether the component has been discarded?

I still prefer the current approach, because it works today, and even with Suspense, it should work in 99% of cases. It would be a shame to hobble performance and complicate the API for a chance Suspense messes with it. I'd still prefer some kind of fix or workaround.

Another thing I'm not sure about is how concurrency plays in to all this. Can a render's `useEffect` hook be cancelled with concurrency just like they are for Suspense? I don't know the answer to that.

CaptainN:
We could probably just offer a second suspense compatible version - `useHighwireTracker` - it would be relatively easier to build that. Just use `useEffect` and pass in deps, and return null for the initial render.

BTW, I merged the `useMemo` version and the computation life-cycle stuff into a new branch, and I'm think to make a PR out of that: https://github.com/CaptainN/react-packages/tree/hooks-plus-ultra

CaptainN:
As far as I can tell, the problem with Suspense is isolated to one specific scenario, one that I think `withTracker` cannot even trigger, and is easily avoidable with `useTracker`. That scenario is if you use the hook before a thrown promise in the hook order. If you use the hook *after* all possible thrown promises (which would always be the case for `withTracker` components, since they are not instantiated typically until the return statement) then you can't trigger the problem - it works fine. If you avoid using `useTracker` along side possible throws altogether (and use nested components), then you just don't have a problem. Am I missing another problematic scenario?

packages/react-meteor-data/useTracker.js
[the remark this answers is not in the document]
CaptainN: Check the new PR #273, this is resolved differently

CaptainN:
> Not sure what would happen for reactive events triggered between the initial render and the actual mount, though. We wouldn't want them to trigger a rerender before useEffect runs and ensure our render was committed ?

@yched Sorry, I didn't see this reply until now. Yes, that is the one gotcha I noticed as well. I think we'd have to accept a gap in service, so to speak in that case, but I think this workaround will do the trick for 99.9% of cases, even without a service gap:

1. In `useMemo` set up computation, and timeout@50ms, also set unmounted flag
2. reactive updates that happen before `useEffect` run `reactiveFn`, but do not try to set react state. Also set a rebuild flag - in other words, if a reactive update happens between initial computation creation, and `useEffect`, we'll do a rebuild in `useEffect`.
3. If the timeout happens, and the computation is disposed, reactive updates will not trigger the `reactiveFn`. In most cases that will not be a problem, but we should document it.
4. In `useEffect`, cancel timeout, and set mounted flag. If timeout timed out, or rebuild flag was set, restart computation, and do another render. This will run the computation, and get us back up to date with the current state of the `reactiveFn`.

Another thing I'd like to explore is what happens if you do try to update state in the computation from a dead end react component - does it throw errors or anything we can catch? Then we wouldn't even need a timeout (a quick test didn't reveal anything, but maybe there is something we can latch on to).

yched:
Posting this here so as not to pollute the other PR too early - look what recently landed as a (for now private) standalone package in the official react repo :

https://github.com/facebook/react/blob/master/packages/use-subscription/README.md

If you read the discussion in the associated PR https://github.com/facebook/react/pull/15022, it's almost laughable how they hit the exact same pain points as we did :-D. Interestingly, they admit they're not able to fully optimize in concurrent mode at the moment.

Their hook over there is designed for an API with clearly separated "subscribe to something" and "receive a value from that subscription" steps, whereas Tracker.autorun mingles the two, so not sure we'll be able to use it directly. But at the very least the code and discussion there are super relevant for us here...

If need be, it could make sense to reach out to the PR author at some point to discuss the specifics of our case, since he now has a pretty good vision of what is or isn't possible at the moment or in the future with concurrent mode.

CaptainN:
I figured out while writing up a comment on React's `useSubscription` that actually `useMemo` with a cleanup method wouldn't work, if it doesn't have a semantic guarantee, because that could stop the computation and prevent some updates from applying. Bummer.

CaptainN:
Reading through what they did, we can't use that. They made an assumption that subscribe is "passive" - which allowed them to avoid setting up any side-effects during render. I think they also assume the side-effects they set up in `useEffect` will not trigger another render until some later time, or they are simply willing to accept the second render.

yched:
Yup, came to the same conclusion, sadly :-/

They split between a "getCurrentValue()" callback that is supposed to be side-effects-free, and a "subscribe()" callback were side-effects occur, so they can can just call getCurrentValue() when needed to get the "sync value for current render". That's not true in our case, the function that gets the value also performs side effects - that's where I monkey-patched DDP subscriptions in my original PR...

So, right, their useSubscription is not for us it seems.

yched:
Also - for the timeout-based approach, you wrote in https://github.com/meteor/react-packages/pull/271#issuecomment-515651974

> reactive updates that happen before useEffect run reactiveFn, but do not try to re-render (do not set state). Also set a deferred render flag - in other words, if a reactive update happens between initial computation creation, and useEffect, **we'll do a deferred render in useEffect**.

Yup, apparently this is exactly what `useSubscription` refer to when they write "it achieves correctness by sometimes de-opting to synchronous mode" in their [README](https://github.com/facebook/react/blob/master/packages/use-subscription/README.md#limitations-in-concurrent-mode) : catching up with missed value changes that happened between initital render and the actual subscription at effect-time, by forcing a re-render directly from within the execution of useEffect() - apparently those renders are treated synchronously. That defeats the benefits of concurrent rendering when that happens, but since it is acceptable for their use case, that should be acceptable for us as well :-)

CaptainN:
So one other thing we should look in to is how to handle error scenarios (error boundaries). I'm concerned that when an error it's thrown we can end up with the same sorts of cases where things are leaking memory and computations are not cleaned up.

CaptainN:
@yched @menelike I think I got a version that satisfies all the requirements. [Take a look.](https://github.com/CaptainN/react-packages/blob/hooks-suspense/packages/react-meteor-data/useTracker.js)

The only part I'm struggling with now is writing some unit tests for the edge cases of concurrent mode (I'm not sure how to get the test thingy to wait for the thrown promise to resolve in my simulated Suspense rigging, but it seems to actually work), here's what it's doing:

1. On first render (unmounted) it will create a timeout@50ms, which will dispose the computation.
2. In `useEffect` it will cancel the timeout. If the timeout has no occurred, it will simply continue using the computation.
3. It will not run `reactiveFn` at all if the component is not mounted, except on `firstRun`, synchronously. This is to reduce overhead for discarded renders, and to keep consistent with the aims of the react team's push for efficiency. It basically only runs the computation once per discarded render, and memory/resources are (essentially) garbage collected after 50ms, outside the main render cycle.
4. If a reactive update does sneak in between first render and `useEffect`, a deferred render flag is set.
5. Also in `useEffect`, it checks for any deferred renders, which should have been triggered by reactive changes, which happened between first render, and `useEffect`.
6. In `useEffect` we also check to make sure the computation has not been discarded - if it has, it starts up a new computation and forces re-render.

After chatting with Brian Vaughn in facebook/react#15022 it became clear that this is an important area to work around in multiple cases - suspense, concurrent mode and other memory optimizations, and error boundaries. Suspense and error boundaries are shipping in some form or another today. What I've basically set up is a poor man's garbage collector based on `setTimeout`.

Once I've gotten the TinyTests to work the way I want, and have updated the readme to explain all this (including a section on Concurrent mode and avoiding side-effects in the reactiveFn) I'll merge this with PR #273.

CaptainN:
So, the way `useEffect` works is different from `componentDidMount` - the former yields to allow the browser to paint, and the latter runs synchronously. It would actually be easier to build all this, with a shorter timeout period using `componentDidMount` because of this, maybe even just 1 millisecond. I don't think it's worth rewriting the `withTracker` component just for that more simple implementation of the same algorithm, but it's an option worth knowing about.

Another thing I thought about is, can we leverage some other browser APIs such as `requestAnimationFrame` to get a faster answer about whether to cleanup (faster than a timeout of 50ms)? I probably won't play with this much, since it would require digging in to react internals, or writing a bunch of tests (or both), but I thought of it, and wanted to document it somewhere.

Issue #242, opened 2017-12-29
componentWillMount vs componentDidMount


crapthings:
https://github.com/meteor/react-packages/blob/devel/packages/react-meteor-data/ReactMeteorData.jsx#L127

componentWillMount vs componentDidMount

what is different between

```javascript
componentWillMount() {
  this.data = {};
  this._meteorDataManager = new MeteorDataManager(this);
  const newData = this._meteorDataManager.calculateData();
  this._meteorDataManager.updateData(newData);
}
```

```javascript
componentDidMount() {
  this.data = {};
  this._meteorDataManager = new MeteorDataManager(this);
  const newData = this._meteorDataManager.calculateData();
  this._meteorDataManager.updateData(newData);
}
```

if we use react-router or flow-router + react-mounter then we make a subscribe withTracker

when switching route, next component's componentWillMount will happen before previous component's componentWillUnmount. is this a correct behavior?

this will cause minimongo's remain previous subscription's records, it reruns component while previous records are removed one by one.

the ddp message is always show
sub first then unsub

![image](https://user-images.githubusercontent.com/1147704/34431454-439920e6-ecaa-11e7-9f7a-5eea08a63c8e.png)

this will cause unwanted rerun, screen flick.

then i try to make a container like what react-meteor-data does then move subscribe to "componentDidMount" and "componentWillUnmount" to stop tracker first and subscribeHandler

![image](https://user-images.githubusercontent.com/1147704/34431491-abe5c1ea-ecaa-11e7-84f9-f9f8a549df8a.png)

![image](https://user-images.githubusercontent.com/1147704/34431519-e3e4ed1e-ecaa-11e7-8c7f-1208f330c2ee.png)

the order looks good now, and stop subscribe does what meteor docs said

![image](https://user-images.githubusercontent.com/1147704/34431568-2bb7a3a2-ecab-11e7-9707-76f32add2450.png)

> stop()
Cancel the subscription. This will typically result in the server directing the client to remove the subscription’s data from the client’s cache.

crapthings:
some issue relative to this.

Clean client side minimongo?
https://github.com/meteor/meteor-feature-requests/issues/242

dburles:
@crapthings PR definitely welcome 😊

filipenevola:
closing because of react-meteor-data@2.0.0 https://github.com/meteor/react-packages/pull/273

PR #273, opened 2019-07-26
Optimized useTracker hook

CaptainN:
This PR is a roundup of the discussion (and builds on the work) in #262 and #271, and supersedes: #242, #252, #256, #263, #266, #267, #268.

It does the following:

- Create a base hook, `useTracker` to for use as an alternative to withTracker in hooks based projects.
Rewrites `withTracker` on top of `useTracker` (issues #256, #252, #242 / PR #261)
- Updates docs!
- Removes deprecated createContainer.
- Above #262, this PR retains the synchronous rendering behavior in `withTracker` and in `useTracker` if no deps are specified, but also allows asynchronous reactivity after `firstRun` if deps are supplied, and properly responds to deps changes. This implementation is [compatible with Concurrent Mode, React Suspense and Error Boundaries](https://github.com/meteor/react-packages/pull/271#issuecomment-518414448).
- I wrote new TinyTests for the hook, and ported back the old ones.
- Above #271, this PR fixes a number of issues raised in that PR.
- It also switches the underlying deps compare functionality from a fork of React internals to using `useRef`. There was some concern that results from that may be randomly tossed, but for now (react 16.8.6) they are stable. Even if they are tossed, the only effect this will have on our implementation is that it'll recreate the computation on the next render.
- This PR also adds a third param to the hook, which is a computation lifecycle hook. This method is structured like the method passed to `useEffect`. The main function runs when a new computation is created (before `reactiveFn`) and then if it returns a function, that function is run right before the computation is destroyed. I don't know if this is really all that useful in real world use cases, but it is useful for testing some internal behaviors. I left it undocumented for now.
- This PR also now passes a reference to the current computation to the `reactiveFn`.
- Compatible with SSR.

Since this bumps the react version requirement to 16.8, there has been some discussion about how to properly release this with a major version bump. One suggestion was to bump the old version to 1.0.0 which can receive updates separately, then release this update as 2.0.0.

I believe this is ready for review. (Third time's a charm!)

CC @hwillson @menelike @yched


CaptainN:
After discussing this with Brian Vaughn in facebook/react#15022 it became clear that it is important to support suspense, concurrent mode and error boundaries, and to avoid memory leaks associated with creating side-effects in the component directly. Suspense and error boundaries are available in React today.

The way I set that up is to use a [poor man's garbage collector](https://github.com/meteor/react-packages/pull/271#issuecomment-518414448) based on setTimeout rather than trying to move all the side-effect creating stuff into `useEffect`, which would not be backward compatible or as efficient in most cases as the approach in this PR.

Now this is really really ready for review.

packages/react-meteor-data/useTracker.js
```diff
@@ -0,0 +1,188 @@
...
+      }
+      runReactiveFn(c);
+    } else {
+      // If deps are anything other than an array, stop computation and let next render handle reactiveFn.
+      // These null and undefined checks are optimizations to avoid calling Array.isArray in these cases.
+      if (deps === null || deps === undefined || !Array.isArray(deps)) {
+        dispose();
+      } else if (refs.isMounted) {
+        // Only run the reactiveFn if the component is mounted.
+        runReactiveFn(c);
```
menelike: can `forceUpdate` be moved into this if clause?

packages/react-meteor-data/useTracker.js
```diff
@@ -0,0 +1,188 @@
...
+      }
+    }
+  }, deps);
+
+  useEffect(() => {
+    // Now that we are mounted, we can set the flag, and cancel the timeout
+    refs.isMounted = true;
+
+    if (!Meteor.isServer) {
+      clearTimeout(refs.disposeId);
```
menelike: minor note, it's cleaner to check `refs.disposeId` and if given `clearTimeout`

packages/react-meteor-data/useTracker.js
CaptainN: The logic is a little different there, for whether to call forceUpdate. I actually had some logic to avoid calling `forceUpdate` for unmounted components, but it didn't cause any problems to just call it in every case, and some of the older tests were failing with the logic in place. So I just left it. I can try it again, but it didn't seem important.

packages/react-meteor-data/useTracker.js
CaptainN: At this point in the execution there will always be a `disposeId`. It always creates a timeout on first render (onBeforeMount), then the `useEffect` runs at some point after that (or never if the render is tossed), so there's always a `disposeId` to cancel.

packages/react-meteor-data/useTracker.js
CaptainN: Actually, it was easier than I thought - that's what I get for coding at 2:00am. This is adjusted as suggested in the latest commit.

menelike:
This looks as perfect as it can get, I'll run some tests today and write back the results!

Very nice job @CaptainN ❤️🚀

CaptainN:
We gotta get this merged, so that I can stop making "improvements" lol

Okay, so after reviewing all the things that I've learned about hooks and `useEffect` and `useLayoutEffect` (and reading some other issues on React, and how others are handling various things) I think we have one more big change we should make. We should change from `useEffect` to `useLayoutEffect`. These two hooks are basically the same thing, except `useLayoutEffect` runs synchronously with the current render (and allows to *optionally* set up second synchronous render). 

In general, it's better to do expensive things in `useEffect` instead of synchronously with render, but we are already doing the expensive things in `useMemo` - `useEffect` is just used for some very quick book keeping. Moving to `useLayoutEffect` will greatly simplify some of the edge case checks we need, and shorten the length of time we need to wait for confirmation that the current render will be committed (I think we can get away with 1 or even 0 milliseconds).

CaptainN:
I'm going to roll back the change to `useLayoutEffect` - I forgot about the time it takes to update the DOM - in concurrent mode it doesn't seem to perform any better than `useEffect` for our purposes, and I removed a lot of the important checks for incorrectly cleaned up renders, which leaves the hook in an nonreactive state (oops!).

Some interesting observations: It turns out that quite a few renders are tossed away in concurrent mode, at least in my test app (roughly 2 tosses to 1 commit). It also can take quite a long time for the commit to land - 50ms actually misses MOST of them. I'll increase the timeout period to 1 second. If it takes longer than that, we'll just have to eat a rerender.

CaptainN:
More discussion for the pile in facebook/react#15317

CaptainN:
Another quick note - I'm using this in my current projects without concurrent mode. I left a `console.log` in place for the case where a computation is cleaned up by the timeout, and then restarted in `useEffect`. In practice, this rarely happens even in concurrent mode (at least for my app - I had actually forgotten the log was even in there). It's even rarer in non-concurrent mode - but it has happened (exactly one time, during a particularly janky reload during development). It's worth knowing that it can happen, that commit after render takes longer than 1 second, even without concurrent mode, and that the hook dutifully restores the reactivity when it needs to.

I think this version is as @menelike said, is "as perfect as it can get." (not much has really changed since that comment 10 days ago). I think it's ready for broader review and testing. @hwillson

CaptainN:
I was just thinking about how this could be more efficient in concurrent mode, and clean up the code a bit (remove timeouts, etc.). If react offered some kind of deterministic ID for each leaf of rendering, we could actually do an algorithm very close to what we have now very cleanly. But they don't offer that.

But what if we ask the user to supply a globally unique name for each `useTracker` usage. Behind the scenes we could then store whatever computation we create in render, by that ID, then if 3 of 4 renders are never committed, that's okay - we'll only have ever created one computation. When the one that does get committed is cleaned up, we could destroy it. We could avoid disposing in-use computations by keeping a "committed" count on each named `useTracker` store

This would be a significant change to the API - is it too onerous to ask users supply a globally unique ID/name for each instance of `useTracker`?

yched:
> is it too onerous to ask users supply a globally unique ID/name for each instance of useTracker?

That sounds fairly tedious, yes ;-)

dburles:
@CaptainN could you not handle that problem in a similar way to how `useState` works? that is by storing a reference within a global array (like a subscriber pattern). Or perhaps I'm missing the finer details of the problem.

menelike:
> I think this version is as @menelike said, is "as perfect as it can get." (not much has really changed since that comment 10 days ago).

I've refactored our large project to `useTracker` (though we're using a variant in https://github.com/risetechnologies/react-packages), it was a satisfying experience removing `withTracker` ;).

One thing I have encountered is that if the result of a computation did not change, a `forceUpdate` is just a waste of computing resources. We're using `transforms` and `isEqual` for more fined gained control on this, but I think that in this code a simple `Object.is` (which should be polyfilled) could achieve some performance improvements at mostly no cost like: 

```
        const d = runReactiveFn(refs, reactiveFn);
        const hasChanged = !isEqual(refs.trackerData, d);
        if (hasChanged) {
          refs.trackerData = d;
          forceUpdate();
        }
````

From my experience, I also encourage every developer to output flat or easily comparable results from computations. We use https://www.npmjs.com/package/shallow-compare to transform objects and arrays on a global/default level so that when no change has been detected the previous computation result is used where on the second step `Object.is` returns true and a re-render is skipped. 

> is it too onerous to ask users supply a globally unique ID/name for each instance of useTracker?

honestly, this would destroy the current nice developer experience.

**Comment by yched** [comment]
> One thing I have encountered is that if the result of a computation did not change, a forceUpdate is just a waste of computing resources

Yes, as I pointed in the previous PR, triggering rerenders with setState rather than forceUpdate would  take care of that, as the built-in useState is already optimized for that. Altough TBH, after all the approaches that were considered since we began, I kind of lost track whether setState can be leveraged with the current "setTimer-based GC" approach.

However, with minimongo returning a different object on each read, that's a fairly large part of the reactive sources that are out of reach for any such optimization anyway... I'm always a bit surprised to find little discussion or interest in an immutable minimongo :-)

**Comment by CaptainN** [comment]
@dburles I wonder if we could do that. The React teams uses some deep internal APIs to implement their hooks, so they have a better way to tie a current hook to a render in ways they don't expose to external developers (grumble,grumble...). But I did do some testing with a counter to try to see what was going on in concurrent mode. I noticed that multiple render passes use the same incremented number, indicating (I guess) that they are running things in web-workers? I'm not sure it was consistent though, I'd have to do a lot more testing. If those incremented values are reliable, we could do it that way. I hesitate to try and do that now because concurrent mode is considered unstable, and I don't want to dive too deeply until it's unlikely to change again. The work we have now is most important for Suspense and Error Bounds.

@menelike I'm okay with a reactive update forcing update whenever it changes, even if nothing changed. The contract with MiniMongo is that we get a reactive update when something changes - so it should be safe to assume something changed. If not, the place to fix something like that is in MiniMongo IMHO.

@yched I think `setState` wouldn't work better than our current solution because of what you mentioned - the data we get back from MiniMongo is not immutable.

One other hair brained idea, what if the requested ID was optional, like `deps`? If supplied, our tool would only create one computation and use it for all multi-renders, and if not, it would work like it does now with `setTimeout`. This would at least allow a user to control cases where they really don't want 4 computations spun up for the one `reactiveFn`. I think out cleanup routine would be much simpler too, since I think we can assume if the render is run 4 times, it'll be committed at least once, so we can just rely on the `useEffect` cleanup function to do its thing.

**Comment by menelike** [comment]
@CaptainN 
I think you're forgetting that minimongo also fires if unrelated fields of the same document have been changed (even if you add the projection to your query options), which from my experience happens a lot of you're not working on a todo app ;).  Also, it would encourage developer to write computations similar to reducers.

My recommendation is based on 4+ years experience with meteor in large projects. Just my two cents.

**Comment by CaptainN** [comment]
@menelike We had discussed elsewhere adding a compare method. I still think that makes sense for this case. The third parameter is currently undocumented (it provides access to the computation lifecycle). We could replace that with an options object which could contain both the computation lifecycle handler, and a custom comparison function (like shallow compare). That would get you exactly what you need. You could even generalize it:

```js
const useMyTracker = (func, deps) => 
  useTracker(func, deps, {
    shouldUpdate: (prev, next) => 
      shallowEqualObjects(prev, next) // or any custom compare
    }
  );
```

menelike:
@CaptainN I absolutely had no intention to make this PR more complex, that's the reason why I suggested a simple builtin compare (at no costs) with `Object.is()`. But again, it is just a suggestion (when I see that there might be a way to increase performance at no cost I think it's worth the pain ;) )

But more importantly, this PR can be merged, all my tests over the past days worked very well!

vladbalan:
Great work guys! And thank you! When can we expect this to be merged?

CaptainN:
@menelike I do like the idea of being able to reach in to the hook to do certain types of optimizations (like being able to leverage the computation lifecycle, and adding a compare function as described). It wouldn't take long to implement a feature like that (and to test it), but I'd love to get some feedback from the powers that be, and get this thing on the road to merged.

CaptainN:
I was curious how concurrent issues (and suspense/error boundaries) was handled in Apollo - it looks like it's not supported in Apollo's `useQuery` hook - [won't this leak memory](https://github.com/apollographql/react-apollo/blob/91719db6239f143ffac656005a9c01b40d2609d6/packages/ssr/src/getDataFromTree.ts#L54)? @hwillson

CaptainN:
Any chance for a review in the near term?

CaptainN:
Check out how much easier it was to write a Svelte hook 😜 :

```js
import { Meteor } from 'meteor/meteor'
import { Tracker } from 'meteor/tracker'
import { onDestroy } from 'svelte'

function trackClient (func) {
  // Use Tracker.nonreactive in case we are inside a Tracker Computation.
  // This can happen if someone starts rendering inside a Computation.
  // In that case, we want to opt out of the normal behavior of nested
  // Computations, where if the outer one is invalidated or stopped,
  // it stops the inner one.
  const computation = Tracker.nonreactive(() =>
    Tracker.autorun(func)
  )
  onDestroy(() => {
    computation.stop()
  })
}
function trackServer (func) {
  return func()
}

export const track = Meteor.isClient
  ? trackClient
  : trackServer
```

Seriously though, there's a lot of tricky code in this PR, and if it would help, I'd be happy to jump in a Slack/Discord chat or a voice call to go through the code and explain why it's the way it is.

CaptainN:
I'm reading through the Tracker code, to see if there is some way to run the reactive function and capture deps, and then set up the computation later in `useEffect` without rerendering.

While looking through, I found out that Tracker runs the reactive function on the sever in a `Meteor._noYieldsAllowed` block. Why is that important, and should our hook be doing that too?

CaptainN:
It looks like calling `Tracker.autorun` with an empty function alone would not create any persistent instances of anything. Is that right?

From what I can tell reading through the code (at 12 at night, heh), since we are running in a nonreactive block, the only real reference to the computation will be held by react in our hook, and by whatever Deps are tied to it using `Dependency.depend()`. It does this by grabbing a reference to `Tracker.currentComputation` (which is set by `Tracker.Computation` in its constructor during `firstRun`), then each dep manages it's own list of computations by computation.id (an incremented integer).

This brings us back to the idea that we can just use an id to make sure all of any given render pass's computations are the same one computation. Then react and render, pause and toss all it wants, and we'll just always have that one computation. It's too bad we don't have a deterministic way to generate an id - we could just stuff that in to the computations's id prop if we had it (or manage it ourselves), and then we would be able to clear out so much of this timeout based hackery.

Could we do something perhaps crazy and radical like `function.toString()` to generate a unique id?

(There are other things I could do to make the code easier to understand - for example, separating the version with deps from the version without and placing a front controller between the two might help with code legibility - I'd be happy to do this if the complexity is holding up review).

CaptainN:
Actually, I was originally thinking of a way to override whatever mechanism captures the computation and holds it in memory in Meteor. That seems to be all in [`Dependency.depend`](https://github.com/meteor/meteor/blob/5c32331591f268f2acdac5a1a7481eb9e65dc3d4/packages/tracker/tracker.js#L396). Could that be replaced with a [WeakSet](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet) instead of a local object? The reasoning being that if the reference were garbage collected because React tossed the instance, that it would get garbage collected from this dispatch list too.

CaptainN:
Aside from the idea above, about leveraging the js GC to clean up discarded computations, I have other thoughts.

One idea is that we basically have two versions of the hook - one with deps, and one without - or put another way, one with an optimization model, and one without. The react team says it's not important to avoid the second render caused by the use of `useEffect`. I find that claim dubious (especially when not using concurrent mode), but let's just accept it as a given for the sake of argument. It would make the implementation much cleaner if we were to accept that.

But there may still be a way to get what we want - an optimized hook, even without all this set timeout complexity. What if the standard hook (no deps) did it the way the React team suggests, and simply renders twice for every mount, but we then provide a set of optimization stages to prevent that for those who are willing to put in that work. There are 2 main optimizations I'm thinking of; `deps` and a `shouldUpdate` method, which could be added for each defined hook built on `useTracker`, which could check the return data and see if anything has actually changed - if not, don't force update.

These two methods could get us what we want - a `useTracker` hook (configured) which doesn't render twice, and absolutely cannot leak memory, because it does avoid creating side effects in render. The big difference is that it would render twice by default, which the current implementation does not. The algorithm for setting all this up would be much easier to read.

Short of that, refactoring the current hook to separate the deps version from the no-deps version might at least help with review (it'd make the code easier to read). It would be nice to get some feedback about any of this, even if it's to say "we're not interested" - in that case I'd just release this on my own, and stop worrying. :-)

StorytellerCZ:
@hwillson @benjamn Can we please get a review and merge on this or at least #274 ? The warnings are flooding the console and making debugging more difficult.

zhaoyao91:
what advantage would this PR have over [react-meteor-hook](https://github.com/andruschka/react-meteor-hooks) pacakge?

StorytellerCZ:
@zhaoyao91 It being in the official package.

CaptainN:
> what advantage would this PR have over [react-meteor-hook](https://github.com/andruschka/react-meteor-hooks) pacakge?

The PR doesn't just add a hook - it also re-implements `withTracker` on top of it, to gain efficiency and support for suspense, error boundaries, and concurrent mode. I don't think even the apollo hook can boast support for those things.

CaptainN:
I have an alternative solution to providing a hook that is compatible with React's "rules of hooks". We could do these things:

1. Write a set of hooks that don't create side effects on render, but only in in `useEffect` and `useLayoutEffect`. Those hooks would be very very simple.
2. Don't bother rewriting `withTracker` on top of `useTracker` - just use the `UNSAFE` prefixed version of `componentWillMount` and move on. The result of a "pure" hook would be incompatible with `withTracker`'s legacy behavior.

```js
function useTracker (fn, deps =[]) {
  const [data, setData] = useState(null)
  useEffect(() => {
    const compputation = Tracker.nonreactive(() =>
      Tracker.autorun((c) => {
         setData(fn(c))
      })
    )
    return () => { computation.stop() }
  }, deps)
}
function useLayoutTracker (fn, deps = []) {
  const [data, setData] = useState(null)
  useLayoutEffect(() => {
    const compputation = Tracker.nonreactive(() =>
      Tracker.autorun((c) => {
         setData(fn(c))
      })
    )
    return () => { computation.stop() }
  }, deps)
}
```

The upside of these hooks is that the implementation is super clean and free from side-effects. The downside is that the user will have to expect a `null` value for at least one render. In many cases this may not be a big deal, since they are probably doing a subscribe and checking `ready()` anyway, but it is something they'll have to be aware of. This makes the implementation not backward compatible with `withTracker` so it can't be used to re-implement `withTracker`.

CaptainN:
I've signed the updated CLA. I think @menelike and @yched must also sign the updated one for this PR.

yched:
Signed the CLA as well.

menelike:
signed as well.

castasamu:
Great work guys!!!

JanMP:
I tried react-meteor-hooks, withTracker and this PR. I had loads of problems with react-meteor-hooks. This PR works like a charm. I would really, really like to keep using this.

markshust:
how can one apply the updates in this PR to a meteor instance without waiting for the new version to be published?

CaptainN:
@markshust You can checkout a copy of [the repo](https://github.com/CaptainN/react-packages/tree/hooks-plus-ultra) and copy the `packages/react-meteor-data/` folder to your project's lcoal `packages` folder. Meteor will then use that version instead of the one in atmosphere.

CaptainN:
I think the CLA is signed by everyone who needs to - not sure why it's not validating.

CaptainN:
I think we've all signed the updated CLA, but it still says no, with no additional information.

menelike:
I've signed like 6 times just to not be the one who was to dumb to agree to the CLA.
I've to admit the signing UX feels like something done in a hackathon.

StorytellerCZ:
If I remember correctly it was done during a hackathon. 😅

CaptainN:
There's an interesting [post in the forums](https://forums.meteor.com/t/handle-publication-errors-in-withtracker/50498) where a user tried to use a hook inside a `withTracker` handler, which is of course not a component directly. This usage is perfectly acceptable after this PR.

Based on the complete lack of any feedback for many months, I guess there is no interest in this PR - should I close it?

robraux:
I'm not sure where to adequately add my +1 and convey just how useful this would be for us. IMO this is something which is core to Meteor's continued support of React.

ni-ko-o-kin:
Hey @CaptainN many thanks for this awesome PR!

Who is responsible for merging this PR?

StorytellerCZ:
I just wrote an e-mail to Tiny to see if anyone can get this merged. Let's give them few days.

dburles:
The hold up is that we require someone with permission to publish it to Atmosphere. To reiterate the plan again, we should first publish the current version as v1.0.0 followed by merging and publishing this as v2.0.0.

CaptainN:
I also recommend silencing the warning by switching to the UNSAFE prefixed lifecycle methods for v1.0.0

benjamn:
@StorytellerCZ @dburles Feel free to ping me if you don't hear back soon (say, this week). Seems good for Tiny to go through the exercise of publishing a core package, but of course I can do it too if need be.

dburles:
> @StorytellerCZ @dburles Feel free to ping me if you don't hear back soon (say, this week). Seems good for Tiny to go through the exercise of publishing a core package, but of course I can do it too if need be.

I think so too. There's a bit to take in here, but we have a fairly good idea on the process to release now so we can help with that side.

@CaptainN I think it would make sense to open a new PR for `v1` (named `v1`) including any changes we feel makes sense to include. We *may* have to maintain a v1.0.0 release track but I don't think it's essential. At the very least the `v1` branch can then serve that purpose if we ever need to publish updates. Thoughts?

CaptainN:
I agree about keeping a v1 branch maintained, and maybe even backward compatible with older react version (somehow detect and use the UNSAFE prefixed for newer versions of react)?

It is a mono repo, but the main grouped tests don't seem to have been wired together for a while, and I've ported the tests into the package directly in this PR, so it's ready to be split out.

CaptainN:
Another [forum post](https://forums.meteor.com/t/is-it-possible-to-access-state-within-withtracker/50529/3) about using state within `withTracker`.

filipenevola:
Hi @CaptainN and all, now that 1.8.2 is published we can start working on publish this new version.

Do we have any pending issues here? Any reported issue in the forums that should be included in this conversation?

Floriferous:
This is very helpful for all of us using React! Awesome :)

Would it be worth it to briefly think of a path forward towards supporting Suspense and concurrent mode with this hook? At least make sure that the current API would not have to be reworked and that it could be a minor release in the future?

CaptainN:
@filipenevola There is nothing outstanding that should prevent a review. There are some additional features I might like to add (specifically an API to specify a "shouldUpdate" method), but nothing that would hold up a review of at least the general approach we've taken here.

CaptainN:
@Floriferous Great care was taken to make sure this hook will work with Suspense, concurrent mode and Error Bounds (time slicing based features) without leaking memory or resources. Supporting Suspense API directly is a bit more of an effort, and probably not suitable for this hook because of it's generic nature. That said, there's no reason someone couldn't build a solution which supports Suspense for `Meteor.subscribe` for example on top of this.

filipenevola:
Version 2.0.0 of react-meteor-data is published 🎉

lightningboss:
@filipenevola It would be awesome to close the corresponding issues linked in the PR description. Thanks a lot for merging! 🎉

filipenevola:
@lightningboss done.

if some of them are still pending we can open again, no problem.

dburles:
Great stuff! We should also now cut a v1 branch and publish it as 1.0.0 too.

yched:
Awesome to see this committed and released, thanks @filipenevola, and congrats @CaptainN !

Just : since this went through so many attempts, approaches and refactorings through 3 successive PRs and 150+ commits in total, it would be best to squash into one commit ?
Right now this PR represents the full first 4 pages on https://github.com/meteor/react-packages/commits

filipenevola:
@dburles I just created the `v1` branch https://github.com/meteor/react-packages/pull/274#issuecomment-559560001

Are you going to use v1? I would like to find someone that is going to use v1 then this person can lead the PR that we are going to merge and publish as 1.0.0

filipenevola:
@yched I personally like to see all the changes then we can understand how we ended up in this state but maybe just me hehe 😉

CaptainN:
There was a lot of discussion in 3 PRs and a lot of commits that lead to the way the current code is. It's actually enough that maybe I should write a summary and store that somewhere.

Most of the work for V1 will be in documentation. There is already aPR to rename the deprecated methods.

filipenevola:
Yes, I already sent a message there also https://github.com/meteor/react-packages/pull/274#issuecomment-559560001

1seck:
Hi there,

do you have a readme files for this?
I mean a clean interface definition?
since it looks like quite some effort u put into this.

regards

CaptainN:
It looks like the meteor guide is still recommending `withTracker` - maybe we should update that. As always, PRs are welcome!

There is a [readme](https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/README.md), and there is a [blog post](https://blog.meteor.com/introducing-usetracker-react-hooks-for-meteor-cb00c16d6222) about `useTracker`.

PR #274, opened 2019-09-11
Rename deprecated lifecycle to support React 16.9

rj-david:
This is just to suppress the warnings in React 16.9 that can greatly affect development. This is a stop-gap solution while waiting for the new withTracker/useTracker implementation in another PR

### DIFF
--- a/packages/react-meteor-data/ReactMeteorData.jsx
+++ b/packages/react-meteor-data/ReactMeteorData.jsx
@@ -124,14 +124,14 @@ class MeteorDataManager {
 }
 
 export const ReactMeteorData = {
-  componentWillMount() {
+  UNSAFE_componentWillMount() {
     this.data = {};
     this._meteorDataManager = new MeteorDataManager(this);
     const newData = this._meteorDataManager.calculateData();
     this._meteorDataManager.updateData(newData);
   },
 
-  componentWillUpdate(nextProps, nextState) {
+  UNSAFE_componentWillUpdate(nextProps, nextState) {
     const saveProps = this.props;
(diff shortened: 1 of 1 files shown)


CaptainN:
The deprecation warning only showed up in 16.9, but I don't know when the UNSAFE_ prefixed versions of these methods were added to React. In any case, this probably means the package's minor version at least should be incremented.

rj-david:
Good point. I think it was in React 16.3 making this change incompatible to any version lower

rj-david:
It was in 16.3
https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

CaptainN:
Is it worth maintaining support with older versions of react? We could select between the UNSAFE prefixed method and the regular method based on the react version string stored in `React.version`.

This PR can be the basis for the V1 branch mentioned in #273.

filipenevola:
hi @rj-david I believe this PR needs to be done against a new branch to create a version 1.0, right?

Are you going to use 1.0? It would be good to have someone that is going to use 1.0 leading this branch.

rj-david:
@filipenevola, yes I believe so. But we are not going to remain at 1.0. We will be migrating all our active apps to 2.0

The only reason I created this PR was from the request of Ben because it looked like the PR by @CaptainN will not be merged soon.

filipenevola:
Hm ok, then I'm going to close this PR, if someone wants to use v1 just create a PR against v1 branch.

filipenevola:
1.0.0 published https://github.com/meteor/react-packages/releases/tag/react-meteor-data%401.0.0

PR #277, opened 2019-11-20
Make module lazy

moberegger:
Simply makes the main module [lazy](https://docs.meteor.com/packages/modules.html#Lazy-loading-modules-from-a-package).

Also closes #265


moberegger:
@filipenevola Not sure if I should also include the version bump in this PR or not.

PR #280, opened 2020-01-04
[WIP] Add support for suspense to useTracker

CaptainN:
This is a quick attempt to enable suspense for `useTracker`. There are a couple of things to note:

- [ ] Suspense works a little differently than we might expect - it doesn't suspend, then later commit. Instead the render is never committed, and then it's rebuilt later, which looks a lot like a tossed render which will never be committed in concurrent mode. The workaround is to do the computation twice, once when it throws a Promise, and then again when the promise resolves. This may have an effect on Meteor sources like `Meteor.subscribe` - I'm not actually sure what will happen in this case. Will the subscription data get abandoned as soon as dispose is called on it's computation? Will the next `Meteor.subscribe` (from the next render after the Promise resolves) reuse the same data, and prevent that from being disposed? I have to test this (or hopefully someone knows).
- [ ] figure out what to do with thrown Promises on the server.
- [ ] write tests
- [ ] Check if there is a better way to detect a Promise
- [ ] Check if there is anything we should do to support Error Boundaries
- [ ] Figure out what would happen when `deps` change and a Promise is thrown long after mount - does suspense work that way?

### DIFF
--- a/packages/react-meteor-data/useTracker.js
+++ b/packages/react-meteor-data/useTracker.js
@@ -119,6 +119,8 @@ function useTrackerClient(reactiveFn, deps, computationHandler) {
     // if we are re-creating the computation, we need to stop the old one.
     dispose(refs);
 
+    let suspended = false;
+    
     // Use Tracker.nonreactive in case we are inside a Tracker Computation.
     // This can happen if someone calls `ReactDOM.render` inside a Computation.
     // In that case, we want to opt out of the normal behavior of nested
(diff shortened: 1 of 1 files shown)


CaptainN:
So this is actually kind of tricky. We can't really allow a truly asynchronous operation to linger when created in render (before the render is committed). The subscription has to always be destroyed. Which means in cases where Suspense is used, we have to "deopt" our attempt to retain the render created computation. This would actually make our hook "pure" anyway.

Sigh, the more I read about this time slicing stuff, the more it's becoming obvious that we have to do that anyway. The alternative is to hook in to whatever React team is doing in `useSubscription` to "deopt" in the way React Concurrency can be deopted, but I actually think that's worse than avoiding side-effects and eating a second render cycle.

CC: @menelike @yched

CaptainN:
Okay, so thinking this through - if we stop the computation before the subscription has loaded, does that actually stop the subscription from loading? I think it does, which means there is effectively no way to do support Suspense and concurrent mode at all (except to deopt like `useSuspense` might do).

Unless we gobble up the thrown error/promise in unmounted render passes. But that seems hacky. This has me thinking about how React decides to deopt (break out of concurrency) in some cases, and whether we can intentionally deopt in cases such as these. Hmm, interesting problem.

More broadly, if this is really how Suspense works, then it would seem it's really only useful for loading code modules, and not for data, but that doesn't seem right...

CaptainN:
Well, there it is, Suspense is [not a data fetching library](https://reactjs.org/docs/concurrent-mode-suspense.html#what-is-suspense-exactly), it's meant for fetching libraries.

PR #281, opened 2020-01-04
[WIP] Add a shouldUpdate option to useTracker

CaptainN:
Change the third parameter (currently undocumented) to accept an option object with two optional nodes - `shouldUpdate` and the already present `computationHandler`. This proposed change would make it possible to reduce rerenders by providing a comparison function, to check old values, with new values.

- [ ] Test assumptions, and do any kind of testing
- [ ] Adjust the computationHandler tests to use the option
- [ ] Add tests for the `shouldUpdate` option
- [ ] Add to Readme documentation

@menelike

### DIFF
--- a/packages/react-meteor-data/useTracker.js
+++ b/packages/react-meteor-data/useTracker.js
@@ -86,8 +86,11 @@ const tracked = (refs, c, forceUpdate) => {
       forceUpdate();
     } else if (refs.isMounted) {
       // Only run the reactiveFn if the component is mounted.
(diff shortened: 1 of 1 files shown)


CaptainN:
I'd actually like to make this an optional 2nd OR 3rd argument. The second argument is a deps array, but I'd like to be able to accept a config object instead (overload style). This might make sense to do after the Typescript conversion lands, since there's now a split in the two versions of the hook - with/without deps, and the overload decision can be cleaning implemented there.

PR #300, opened 2020-10-26
Fix #299

StorytellerCZ:
Fix #299 by extending compatible versions. There is no issue with React 17 as there are no breaking changes for that version.

PR #301, opened 2020-10-31
fix(react-template-helper): stop warning about React 17

hexsprite:


filipenevola:
Solved here https://github.com/meteor/react-packages/pull/300 and already released.

PR #303, opened 2020-11-03
Include false in Typescript factory return definition

rijk:
May I suggest we add `false` as well? It already works (the code does a falsy check), and in some cases it allows for cleaner code — up to the user's preference of course:

```ts
useSubscription(() => boardId && Meteor.subscribe('board', boardId), [boardId])
```

You don't have to use it like this, but as the code allows it I see no reason for Typescript to give a warning. Hope you'll consider. Thanks!

### DIFF
--- a/packages/react-mongo/README.md
+++ b/packages/react-mongo/README.md
@@ -18,7 +18,7 @@ Both methods accept a factory method, and deps.
 
 ## useSubscription(factory, deps)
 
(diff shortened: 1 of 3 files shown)


rijk:
Rebased on your latest push.

CaptainN:
I'm not apposed to this, but maybe `null` instead?

CaptainN:
I think this has been superseded by other decisions.

PR #304, opened 2020-11-03
Refactor reactive ready approach

rijk:
- Fixed an infinite loop when calling `handle.ready()`
- Renamed the subscription handle ref to `subscription` to avoid confusion
- Had to move the handle out of the ref, otherwise Typescript got confused by the circular reference (`stop` and `ready` use `ref.subscription`) and you lose typings for `ref`
- Similarly, I had to change the `?.stop()` and `?.ready()` to full checks to silence Typescript warnings

### DIFF
--- a/packages/react-mongo/react-mongo.ts
+++ b/packages/react-mongo/react-mongo.ts
@@ -6,23 +6,16 @@ import { useEffect, useMemo, useReducer, useRef, DependencyList } from 'react'
 const fur = (x: number): number => x + 1
 const useForceUpdate = () => useReducer(fur, 0)[1]
 
-const useSubscriptionClient = (factory: () => Meteor.SubscriptionHandle | void, deps: DependencyList= []) => {
+const useSubscriptionClient = (
+  factory: () => Meteor.SubscriptionHandle | void,
+  deps: DependencyList = []
+): Meteor.SubscriptionHandle => {
   const forceUpdate = useForceUpdate()
   const { current: refs } = useRef<{
-    handle?: Meteor.SubscriptionHandle,
+    /** Automatically triggered if the user called handle.ready() */
     updateOnReady: boolean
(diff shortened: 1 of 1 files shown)


packages/react-mongo/react-mongo.ts
```diff
@@ -32,9 +25,9 @@ const useSubscriptionClient = (factory: () => Meteor.SubscriptionHandle | void,
     // it stops the inner one.
     const computation = Tracker.nonreactive(() => (
       Tracker.autorun(() => {
-        refs.handle = factory()
-        if (!refs.handle) return
-        if (refs.updateOnReady && refs.handle.ready()) {
+        refs.subscription = factory()
+        if (!refs.subscription) return
+        if (refs.updateOnReady && refs.subscription.ready()) {
```
yched, line 30: does that really work ?
updateOnReady starts as false, so the tracker computation doesn't run subscription.ready() and thus doesn't register it as a reactive dep.
Then if the user calls the ready() func we give him, updateOnReady is changed to true, but that doesn't make the tracker computation rerun and become reactive to the ready state change ?

rijk: Ah, good point. I actually swapped these around (`if (refs.subscription.ready() && refs.updateOnReady)`) to fix this, but I took it out of this PR as to not change too much. Do you think that would work if it is swapped?

yched: yep, sounds like that should solve it.
the "become reactive once you actually read it" pattern is not super intuitive IMO :-D 
as discussed in the main PR, I tend to think leaving ready(reactivity out of the hook might be simpler...

rijk: Agreed. I'm working on a PR now, will let you know when it's ready so you can review.

PR #298, opened 2020-10-21
[WIP] New hooks

CaptainN:
A set of packages with new alternative, and much more streamlined hooks for integration between React and Meteor

| Hooks | Package Name
| ------- | ----------------
| `useSubscription` and `useCursor` | `meteor/react-mongo`
| `useMeteorState` | `meteor/react-meteor-state`
| `useUser` and `useUserId` | `meteor/react-accounts`

Basic implementations are in.

Todo:
- Flesh out README.md for each
- Add tests
- Add support for SSR - possibly, some generic SSR helpers
- Maybe add a Method call hook, with SSR helper integration

PR #305, opened 2020-11-03
Attempt to make the useSubscription hook simpler by removing reactivity

rijk:
This is a concept, work in progress based on @yched's comments [here](https://github.com/meteor/react-packages/pull/298#discussion_r516541716) and [here](https://github.com/meteor/react-packages/pull/298#discussion_r516519599).

This changes the `useSubscription` signature to match `Meteor.subscribe`. To be used as follows:

**Normal:**
```ts
const handle = useSubscription('board.projects', boardId)
```

**Conditional:**
```ts
const handle = useSubscription(!!boardId && 'board.projects', boardId)
```

This branch also contains a fix/workaround for the issue that the previous subscription's status is returned for one render after changing the arguments: e37ba55e0507ad1b80afd3fcfa6e3a551fe4d1a3.

### DIFF
--- a/packages/react-mongo/react-mongo.ts
+++ b/packages/react-mongo/react-mongo.ts
@@ -1,30 +1,22 @@
 import { Meteor } from 'meteor/meteor'
 import { Mongo } from 'meteor/mongo'
 import { Tracker } from 'meteor/tracker'
-import { useEffect, useMemo, useReducer, useRef, DependencyList } from 'react'
+import { useEffect, useMemo, useReducer, useState, DependencyList } from 'react'
 
 const fur = (x: number): number => x + 1
 const useForceUpdate = () => useReducer(fur, 0)[1]
 
-const useSubscriptionClient = (factory: () => Meteor.SubscriptionHandle | void, deps: DependencyList= []) => {
-  const forceUpdate = useForceUpdate()
-  const { current: refs } = useRef<{
-    handle?: Meteor.SubscriptionHandle,
-    updateOnReady: boolean
-  }>({
-    handle: {
-      stop () {
-        refs.handle?.stop()
-      },
-      ready () {
-        refs.updateOnReady = true
-        return refs.handle?.ready()
-      }
-    },
-    updateOnReady: false
-  })
+const useSubscriptionClient = (
+  name: string | false,
+  args: any[]
+): void | Meteor.SubscriptionHandle => {
+  const [subscription, setSubscription] = useState<Meteor.SubscriptionHandle>()
 
   useEffect(() => {
+    if (!name) {
+      return setSubscription( null )
+    }
+
     // Use Tracker.nonreactive in case we are inside a Tracker Computation.
     // This can happen if someone calls `ReactDOM.render` inside a Computation.
     // In that case, we want to opt out of the normal behavior of nested
@@ -32,20 +24,14 @@ const useSubscriptionClient = (factory: () => Meteor.SubscriptionHandle | void, 
     // it stops the inner one.
     const computation = Tracker.nonreactive(() => (
       Tracker.autorun(() => {
-        refs.handle = factory()
-        if (!refs.handle) return
-        if (refs.updateOnReady && refs.handle.ready()) {
-          forceUpdate()
-        }
+        setSubscription( Meteor.subscribe( name, ...args ) )
       })
     ))
 
-    return () => {
-      computation.stop()
-    }
-  }, deps)
+    return () => computation.stop()
+  }, [name, ...args])
 
-  return refs.handle
+  return subscription
 }
 
 const useSubscriptionServer = (): Meteor.SubscriptionHandle => ({
@@ -53,10 +39,10 @@ const useSubscriptionServer = (): Meteor.SubscriptionHandle => ({
   ready() { return true }
 })
 
-export const useSubscription = (factory: () => Meteor.SubscriptionHandle | void, deps: DependencyList = []) => (
+export const useSubscription = (name: string | false, ...args: any[]) => (
   Meteor.isServer
     ? useSubscriptionServer()
-    : useSubscriptionClient(factory, deps)
+    : useSubscriptionClient(name, args)
 )
 
 const useCursorClient = <T = any>(factory: () => Mongo.Cursor<T>, deps: DependencyList = []) => {
--- a/packages/react-mongo/react-mongo.ts
+++ b/packages/react-mongo/react-mongo.ts
@@ -33,6 +33,8 @@ const useSubscriptionClient = (
     const computation = Tracker.nonreactive(() => (
       Tracker.autorun(() => {
         subscription.current = Meteor.subscribe( name, ...args )
+        // @ts-ignore this is just an internal thing
+        subscription.current.deps = { name, args }
       })
     ))
 
@@ -41,7 +43,17 @@ const useSubscriptionClient = (
 
   return useCallback(
     () => {
-      return subscription.current?.ready()
+      if (subscription.current) {
+        // @ts-ignore
+        const { deps } = subscription.current
+        if (deps.name === name && deps.args === args) {
+          return subscription.current.ready()
+        } else {
+          // Prevented returning the previous subscription's status
+        }
+      }
+
+      return false
     },
     [name, ...args]
   )
--- a/packages/react-mongo/react-mongo.ts
+++ b/packages/react-mongo/react-mongo.ts
@@ -1,7 +1,14 @@
 import { Meteor } from 'meteor/meteor'
 import { Mongo } from 'meteor/mongo'
 import { Tracker } from 'meteor/tracker'
-import { useEffect, useMemo, useReducer, useState, DependencyList } from 'react'
+import {
+  useEffect,
+  useMemo,
+  useReducer,
+  useRef,
+  useCallback,
+  DependencyList,
+} from 'react'
 
 const fur = (x: number): number => x + 1
 const useForceUpdate = () => useReducer(fur, 0)[1]
@@ -9,12 +16,13 @@ const useForceUpdate = () => useReducer(fur, 0)[1]
 const useSubscriptionClient = (
   name: string | false,
   args: any[]
-): void | Meteor.SubscriptionHandle => {
-  const [subscription, setSubscription] = useState<Meteor.SubscriptionHandle>()
+) => {
+  const subscription = useRef<Meteor.SubscriptionHandle>()
 
   useEffect(() => {
     if (!name) {
-      return setSubscription( null )
+      subscription.current = null
+      return
     }
 
     // Use Tracker.nonreactive in case we are inside a Tracker Computation.
@@ -24,14 +32,19 @@ const useSubscriptionClient = (
     // it stops the inner one.
     const computation = Tracker.nonreactive(() => (
       Tracker.autorun(() => {
-        setSubscription( Meteor.subscribe( name, ...args ) )
+        subscription.current = Meteor.subscribe( name, ...args )
       })
     ))
 
     return () => computation.stop()
   }, [name, ...args])
 
-  return subscription
+  return useCallback(
+    () => {
+      return subscription.current?.ready()
+    },
+    [name, ...args]
+  )
 }
 
 const useSubscriptionServer = (): Meteor.SubscriptionHandle => ({
--- a/packages/react-mongo/types/react-mongo.d.ts
+++ b/packages/react-mongo/types/react-mongo.d.ts
@@ -1,10 +1,9 @@
-import { Meteor } from 'meteor/meteor';
 import { Mongo } from 'meteor/mongo';
 import { DependencyList } from 'react';
 declare type UseSubscriptionOptions = {
     deps?: DependencyList;
     updateOnReady?: boolean;
 };
-export declare const useSubscription: (factory: () => Meteor.SubscriptionHandle | void, deps?: DependencyList | UseSubscriptionOptions) => void;
+export declare const useSubscription: (name: string | false, args: any[]) => () => boolean;
 export declare const useCursor: <T = any>(factory: () => Mongo.Cursor<T>, deps?: DependencyList) => Mongo.Cursor<T>;
 export {};


rijk:
Note that this doesn't work yet; it triggers an error when changing `args`:

image

rijk:
@yched would love your review on this if you have time.

packages/react-mongo/react-mongo.ts
```diff
@@ -1,62 +1,48 @@
...
-      ready () {
-        refs.updateOnReady = true
-        return refs.handle?.ready()
-      }
-    },
-    updateOnReady: false
-  })
+const useSubscriptionClient = (
+  name: string | false,
+  args: any[]
```
yched, line 11: if possible I'd think it would be a good idea to mimick Meteor.subscribe()'s signature : 
useSubscription(name, ...args) 
rather than 
useSubscription(name, single_array_of_args) ?

rijk: Thanks, I did that on `useSubscription`, which is the user-facing API. If I did that here as well, I could no longer destructure to use it as `useEffect` deps.

packages/react-mongo/react-mongo.ts
```diff
@@ -1,62 +1,48 @@
...
 
-    return () => {
-      computation.stop()
-    }
-  }, deps)
+    return () => computation.stop()
+  }, [name, ...args])
 
-  return refs.handle
+  return subscription
```
yched, line 34: The tricky bit is to return a consistent / predictable object to the caller, on the first render, on subsequent renders, on renders when we didn't subscribe because the args were absent :

if we return the subscription object, then we expect consumers to be able to use its .ready() getter, then we must always return something with a (possibly stubbed) .ready() getter.

Or we return the ready() getter func directly (no need to hand the stop() callback to the outside world), but in this case we must always return a function

rijk: That makes sense, I'll change it.

packages/react-mongo/react-mongo.ts
```diff
@@ -1,62 +1,48 @@
...
-        return refs.handle?.ready()
-      }
-    },
-    updateOnReady: false
-  })
+const useSubscriptionClient = (
+  name: string | false,
+  args: any[]
+): void | Meteor.SubscriptionHandle => {
+  const [subscription, setSubscription] = useState<Meteor.SubscriptionHandle>()
```
yched, line 13: I don't think we should put that in state : we don't need to rerender when the subscription changes
(the fact that the subscription changes will probably cause other rerenders because of minimongo results changing, but that is taken care of outside of here)

A useRef should be enough ?

rijk: Hm, that's a really good point. We don't need a rerender for a changing subscription. The ready state might change, though. I'll change it to a ref and see if I can get the ready state to invalidate immediately as well.

yched: the ready state might change, that is handled by the ready() func being a reactive source.
So the consumer can decide to rerender (if he wants) by doing 
const isReady = useTracker(ready)
with the ready function we return ?

yched: ah yes, that means the stub ready() getter we initially return must be reactive too :-/
Well, that might require playing with Tracker deps to build our own reactve func wrapping the one in the subscription handler, rather than handing out that one directly

rijk: I incorporated the changes in 76943f7d144c6fa773cfe18e16f526b150c619dd.

The problem with ready(), is that it will always lag behind one render. So when changing the deps, it will still return the previous subscription's status for one render. I have a suggestion on how to fix that, I'll push that in the next commit.

rijk: Ok so here is a suggested fix for this problem: e37ba55e0507ad1b80afd3fcfa6e3a551fe4d1a3. @CaptainN @yched open to better suggestions 😬

packages/react-mongo/react-mongo.ts
```diff
@@ -24,14 +32,31 @@ const useSubscriptionClient = (
...
     return () => computation.stop()
   }, [name, ...args])
 
-  return subscription
+  return useCallback(
+    () => {
+      if (subscription.current) {
+        // @ts-ignore
+        const { deps } = subscription.current
+        if (deps.name === name && deps.args === args) {
```
yched, line 49: I think you don't need to manually track and compare deps here, you can just set subscription.current back to null in the effect cleanup phase ?

rijk: Well, the thing is, this all happens _after_ the render. So it will already be too late by then.

CaptainN:
I tried a bunch of iterations with this API before settling on a Factory method. There were just too many tricky issues to deal with, and I didn't want to start doing things like deep array compares. One unresolvable issue is that if you have an object based publication API, you are basically not going to be able to manage deps effectively this way, without requiring the implementor to jump through hoops to keep their refs immutable.

packages/react-mongo/react-mongo.ts
yched: The cleanup phase runs after the render, but before the ready() callback you return gets executed I think ?
Then by the time it gets executed, subscription.current has gone back to null and you can just return false ?
(not 100% sure off the top of my head, but worth a try I think)

More importantly, this code still has the same drawback as before : by not running the reactive subscription.current.ready() in some cases, the function we build here won't be reactive later ?

I have something in mind, need to try it first :-)

rijk: > The cleanup phase runs after the render, but before the ready() callback you return gets executed I think ?
Then by the time it gets executed, subscription.current has gone back to null and you can just return false ?

Nope. Tried 10 different ways, even with `useState`, but every time the first `ready()` call still references the old subscription.

rijk:
@CaptainN there's a couple other ideas in this PR as well, if you have time to take a look (in fact, switching out the Factory method was not even the main point).

CaptainN:
It looks like you did a few different things here. One is that the useSubscriptionClient function no longer returns a SubscriptionHandle, but just a single method (the useCallback method). I'm not apposed to that, but it still. has the problem of inadequate deps checks.

yched:
Right, publications can receive args not just as litterals but also as objects or arrays, and we cant use those directly as hook deps (`[] !== []`). That's a bummer, cause the factory API is way less nice :grin:

CaptainN:
I completely agree - this API would be nicer. I just can't think of a way to make it work with inline object refs in a clean way.

rijk:
In terms of usage, I don't see much difference than using `useTracker` though:

```ts
useTracker( () => Meteor.subscribe( 'board', params.id ), [ params.id ] )
// vs
useSubscription( () => Meteor.subscribe( 'board', params.id ), [ params.id ] )
```

What is the difference, or is it just that it allows for a more future-proof implementation?

CaptainN:
It's precisely intended to allow for more future proof implementation, with much easier maintenance. But if you expand the example, it looks pretty different.

```jsx
const MyComponent = ({ slug }) => {
  const sub = useSubscription(() => Meteor.subscribe('board',  slug), [slug])
  const cursor = useCursor(() => MyCollection.find({ slug }), [slug])

  if (sub.ready()) {
    return <Loading />
  } else {
    return <ul>
      {cursor.map(thing => <li key={thing.key}>
        {thing.text}
      </li>}
    </ul>
  }
}
```


```jsx
const MyComponent = ({ slug }) => {
  const { isReady, data } = useTracker(() => {
    const handle = Meteor.subscribe('board',  slug), [slug])
    const data = MyCollection.find({ slug }).fetch()
    return {
      isReady: handle.ready(),
      data: data
    }
  }, [slug])

  if (isReady) {
    return <Loading />
  } else {
    return <ul>
      {data.map(thing => <li key={thing.key}>
        {thing.text}
      </li>}
    </ul>
  }
}
```

The first example gets even simpler if you don't use `.ready()` or you could put the subscribe higher up on the tree, and use more specific queries lower down. You could do that with `useTracker` of course but I think it looks cleaner this way.

I'm actually running in to familiar problems with the useCursor hook in the implementation though. I'm going to have to look at some solutions like `useMutableSource` - which looks precisely like what we need. Sadly, it's in prerelease...

rijk:
@CaptainN I've been toying around with this for the last days, and there is one problem that I keep coming back to:

> The problem with `ready()`, is that it will always lag behind one render. So when changing the deps, it will still return the previous subscription's status for one render.

**I think means this hook needs to manage its own state** as opposed to relying on `useEffect`, as this will always lag one render behind a change in deps. There is no way to immediately clear the subscription, regardless if it's in state or ref.

However, that's not all bad, as it will also allow an out for this issue:

> There were just too many tricky issues to deal with, and I didn't want to start doing things like deep array compares. One unresolvable issue is that if you have an object based publication API, you are basically not going to be able to manage deps effectively this way, without requiring the implementor to jump through hoops to keep their refs immutable.

If we manage our state ourselves, we can use any function we like for comparing `deps`. We can use something like [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) by default, and even give the user the option to pass his own:

```ts
const subscription = useSubscription(
  'page.content', // subscription name (no factory)
  page, // full object passed as argument
  ( a, b ) => a.id === b.id && a.version === b.version // custom compare function
)
```

Or using an off the shelf compare package:

```ts
import equal from 'fast-deep-equal'
const subscription = useSubscription('page.content', page, equal)
```

WDYT?

yched:
Thinking a bit more about the much nicer "non-factory" approach and the question of non-litteral args : 
That's something the original Meteor.subscribe() code already had to solve, since it it _designed_ to be executed multiple times (typically in a reactive function) and do nothing if the args haven't changed. It thus relies on an internal comparison method that can handle object and array params. 
(namely, in [ddp-client/common/livedata_connection.js](https://github.com/meteor/meteor/blob/ab0affe98bb917d26338d96d1ea3520a6dc5cbff/packages/ddp-client/common/livedata_connection.js#L367) - that's based on EJSON.equals()

In this approach where useSubcription() would be a replacement/wrapper for Meteor.subscribe(), it seems totally reasonable that it would rely on EJSON.equals()-based comparisons, just like Meteor.subscribe() does.

packages/react-mongo/react-mongo.ts
CaptainN: fwiw, my other implementation already returns a stub, and takes care of one specific tricky case, which is when React changes props on a mounted component to recycle it, without doing a full lifecycle (react does this often in lists)

CaptainN:
Maybe it's a personal style thing, but I try to avoid leveraging compare options other than those build in to React's core hooks. In case they change their implementation, and to keep dependencies light.

That said, we can use `useMemo` which runs synchronously with render, and accepts deps to get immediate changes. I'll implement a solution based on that.

CaptainN:
@yched That's a great find. I wonder if this has positive implications for actually creating these (and other computations) in render... It looks at a glance similar to an approach I was toying with for a bit.

yched:
This seems to work for me :

```js
const useSubscription = (name, ...args) => {
  const subscription = useRef();
  const params = useRef();
  const readyDep = useRef(new Tracker.Dependency());

  const equalsCommitted = (newName, newArgs) =>
    params.current && params.current.name === newName && EJSON.equals(params.current.args, newArgs);

  useEffect(() => {
    const computation = Tracker.nonreactive(() => (
      Tracker.autorun(() => {
        subscription.current = name ? Meteor.subscribe(name, ...args) : null;
      })
    ));
    if (!equalsCommitted(name, args)) {
      readyDep.current.changed();
    }
    params.current = { name, args };

    return () => computation.stop();
  }, [name, ...args]);

  if (subscription.current && equalsCommitted(name, args)) {
    return subscription.current.ready;
  }
  // Return a fake (but reactive) ready() callback.
  return () => {
    readyDep.current.depend();
    return false;
  };
};
```
Demo code (toggling between no subcription / subscription 1 / subscription 2)
```js
function App() {
  const [value, setValue] = useState(0);
  const ready = useSubscription(value > 0 && 'test_publication', value);
  const isReady = useTracker(ready);
  return (
    <div>
      {value > 0 ? `Subscription with ${value}` : 'No subscription'} - {isReady ? 'ready' : 'not ready'}<br/>
      <button onClick={() => setValue(prev => (prev + 1) % 3)}>Change</button>
    </div>
  );
}
// server-side
Meteor.publish('test_publication', function (param) {
  Meteor._sleepForMs(1000);
  return this.ready();
}); 
```

rijk:
This looks good @yched , I tested and it worked for me as well. Moreover, with this line `const isReady = useTracker(ready)` you leave it to the user if he cares about ready state or not, which could save a couple renders. I used it in a one-liner like this; otherwise the variable names (`ready` / `isReady`) are a little weird IMO.

```ts
const ready = useTracker( useSubscription( 'board', params.id ) )
```

Downside of this is of course we still need `useTracker`!

CaptainN:
@yched That looks pretty good! I don't think you need to worry about managing deps with `equalsCommitted` though, since that's already managed within the Meteor.subscribe function. I think that means we can essentially not worry about constantly rerunning the full computation creation/destruction lifecycle and calling Meteor.subscribe many times.

I also think we can return a ready method that will enable reactivity, rather than relying on useTracker (and a second computation). We already have reactivity from the computation created in useEffect, so we might as well use that.

Your example adjusted:

```jsx
function App() {
  const [value, setValue] = useState(0);
  const subscription = useSubscription(value > 0 && 'test_publication', value);
  const isReady = subscription.ready();
  return (
    <div>
      {value > 0 ? `Subscription with ${value}` : 'No subscription'} - {isReady ? 'ready' : 'not ready'}<br/>
      <button onClick={() => setValue(prev => (prev + 1) % 3)}>Change</button>
    </div>
  );
}
// server-side
Meteor.publish('test_publication', function (param) {
  Meteor._sleepForMs(1000);
  return this.ready();
}); 
```

CaptainN:
I have a version of the above working, but what I'm finding is that I'm jumping through a lot of hoops to avoid 1 re-render when `ready()` changes, and I'm questioning whether it's worth it all to avoid 1 possibly unneeded re-render. Especially since it'll probably coincide with a forced render from `useCursor` anyway. What do you think?

CaptainN:
I pushed a version of this to my branch. Check it out and let me know what you think!

rijk:
Tried it out, and it seems to be working well. Already loving this API:

```ts
// useSubscription( () => Meteor.subscribe( 'boards' ), [] )
useSubscription( 'boards' ) // ✨
```

And coincidentally I also found this makes my Suspense version a perfect drop-in replacement! Not sure if this was intended, but a nice bonus.

```ts
import { useSubscription } from '/lib/react-mongo' // No suspense
// import useSubscription from 'meteor-subscribe-suspense' // Suspense

useSubscription( 'board.projects', boardId )
```

Issue #313, opened 2021-01-23
react-meteor-data v2.2.1 - unnecessary renders


robwheatley:
Perhaps I'm missing something here, but I'm having a spot of bother with v2.2.1. I'm new to react, Javascript and Meteor, so please don't bite!

I recently updated my project to Meteor 2.0. After seeing all the warnings in the logs from withTracker RE undefined deps array, I separately updated react-meteor-data to 2.2.1 as the warnings were due to a bug that was fixed in 2.2.1

I'm using useTracker to tell me when a Meteor subscription is ready, like this:

```
  const projectsReady = useTracker(() => {
    const subscription = Meteor.subscribe('projects')
    return subscription.ready()
  }, [])
```

But when using Meteor Dev tools to watch DDP messages, I see that there are 2 'Subscribing to projects' going off to the server. I don't really see why as I have an empty dependancy array rather than no dependancies. Sure enough, logging on the server shows me that 2 subscription requests arrived there. As a note, the subscription handles I get back had different IDs, but I thought that Tracker.autorun took care of that behind the scenes (looking at Meteor Docs)?

Just to be silly, I did this to see what would happen:

```
  useTracker(() => {
    console.log('Hello world');
  },[])
```

 And, I get 'Hello world' in my log twice (the function is not rendering due to a props change, BTW).

Final note: this didn't used to happen in version 2.1.1 (98% sure of that).

My concern is is two fold - First the extra Subscribes (and corresponding unsubscribes that happen right after) and also the unnecessary renders this causes.

Am I missing something? Is it a bug? Something else?

Looking forward to an expert opinion.

CaptainN:
Subscriptions are a little bit of black magic in Meteor. One of the things that changed in 2.2.1 is that I removed a lot of code which was designed to keep the initial computation (a side effect), created during render alive after the component mounts. This was causing some rare edge case failures with Subscriptions, but it was also making the code base super complex, and difficult to maintain.

All that was replaced with a system which runs and disposes of the computation on render (to avoid creating lingering side effects) and then reruns the computation on mount (in useEffect).

That said, Meteor's subscription black magic should be retaining the data loaded in the initial subscribe call even though it subscribes again, without causing all that data to load a second time. It should be recycling the subscription, or at least the data in the subscription. If that's not happening, we may need to take another look at this.

robwheatley:
Thanks for the reply. I'll check out that PR. In the mean time, I have a little more info to add that may influence things a little..

[...]

PR #314, opened 2021-01-28
Fix lost reactivity when using deps

CaptainN:
This attempts to fix various lost reactivity reports, as well as some trouble with a particular "Maximum update depth exceeded" error, coming from (possibly) constantly updating `deps` interaction with `useEffect`, in two ways:

- Adds guards against containing computations effecting the enclosed computation in `useEffect`.
- Adds a system to keep the reference to `reactiveFn` always up to date with the latest render.

All tests are passing, but I'd like to see if we can work out a test to confirm the reported error.

#313

### DIFF
--- a/packages/react-meteor-data/useTracker.tests.js
+++ b/packages/react-meteor-data/useTracker.tests.js
@@ -220,41 +220,6 @@ if (Meteor.isClient) {
     completed();
   });
 
-  Tinytest.addAsync('useTracker - basic track', async function (test, completed) {
-    var container = document.createElement("DIV");
-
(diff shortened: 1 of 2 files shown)


CaptainN:
We have at least one validation that the approach in this PR fixes the reported issue. @filipenevola Due to the severity of this bug, I don't think we should wait for a test - we should release a new version `2.2.2`.

aureooms:
Repeating what I said in #306:
  - @CaptainN I confirm using `deps` in `useTracker` works as expected in my project with `react-meteor-data@2.2.2`.
  - Could a hook waterfall be used as regression test for this fix?

Issue #317, opened 2021-02-05
Setting state within useTracker causes effects to run reactively


edemaine:
Consider the following code:
```js
defaultPage = new ReactiveVar();
function App() {
  [page, setPage] = useState();
  useEffect(() => {
    console.log(Tracker.currentComputation);
    const observer = Objects.find({page: page}).observe({added: ...});
    return () => observer.stop();
  }, [page]);
  useTracker(() => {
    if (!page) {
      setPage(defaultPage.get());
    }
  }, [!page]);
  ...
}
```

I find that the `useEffect` callback is run within a Tracker (the `console.log` outputs an actual computation), presumably triggered by `setPage` called within `useTracker`.  I tried wrapping `setPage` within `Tracker.nonreactive`, but the effect callback is still called reactively.  This seems wrong...

The result of this is pretty annoying. The tracker actually gets called twice, and because of the `if`, only the first one triggers the observer to be created. Because the observer was created within a tracker, the second call of the tracker invalidates the first call so the observer gets destroyed and not recreated. 🙁 

I tested in `react-meteor-data` 2.1.4, 2.2.1, and 2.2.2 (wondering whether the recent rewrite had an effect) and both have the same issue.  I'm using React 17.0.1

CaptainN:
I don't think you should be mixing APIs in this way. Maybe we should make that clearer in the docs.

Is there a reason you can't just return the data you want from useTracker? Why are you calling `setPage` from in there?

Also, your `useEffect` example has no deps set, which means it'll run after ever render. You probably didn't intend that.

edemaine:
The example is a simplified illustration of a much more complex app... The motivation is that `useEffect` is useful for waiting for the DOM to render (whereas `useTracker`'s first run is before).  In [my actual application](https://github.com/edemaine/cocreate/blob/d3a94e629cee867980cad784844ec89364e592f6/client/DrawApp.coffee#L74-L84), I want to instantiate a complex class that needs both a pageId and DOM refs, so this seemed like a natural solution.  That class happens to build an observer.

In the spirit of not mixing interfaces, I guess I could replace the `useState` with a `ReactiveVar` and a `useTracker` to read it, though, and pass that into an effect.  That would be a nicer workaround than my current workaround (wrapping the `useEffect` callback in `Tracker.nonreactive`, which is pretty confusing because it's not naturally reactive).

[...]

PR #320, opened 2021-02-07
Add warning about setState within useTracker

edemaine:
One approach to resolving #317. Feel free to suggest rewording, or rewrite it yourself. It's a bit of a tricky issue to write...

I hesitate to say "don't mix React and Meteor state", because it *is* fine to set Meteor state (e.g. `Session.set`) within a React hook (e.g. `useEffect`).  It's just the other way around that seems problematic.

### DIFF
--- a/packages/react-meteor-data/README.md
+++ b/packages/react-meteor-data/README.md
@@ -87,7 +87,7 @@ function Foo({ listId }) {
 "react-hooks/exhaustive-deps": ["warn", { "additionalHooks": "useTracker|useSomeOtherHook|..." }]
 ```
 
-**Warning:** It's best not to call React functions that may cause rerenders, such as the `setState` half of a `useState` hook, within `reactiveFn` (or other functions run with `Tracker.autorun`). Doing so can have unintended interactions, where a rerender (including effects in `useEffect`) may execute within a `Tracker`, causing them to invalidate when the original `useTracker` does.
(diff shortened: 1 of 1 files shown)


CaptainN:
That's great! A suggestions to return the state from useTracker, and deal with mutations to that state in the render body, may be called for here as well.

edemaine:
@CaptainN Just took a stab at this. React recommends/requires making state changes only within effects, so instead of "render body" I mentioned `useLayoutEffect` (which is perhaps more consistent with `useTracker`) and `useEffect`.

filipenevola:
@CaptainN could you review one last time?

CaptainN:
It's not exactly right - I'll try to make some more specific suggestions in the coming weeks. The main issue is mixing state managers. It's not really about rerenders (useTracker handles rerenders just fine).

StorytellerCZ:
@CaptainN any update on this?

CaptainN:
I'll review this soon

CaptainN:
I did add a note encouraging users to return data, instead of setting state. But I also fixed an issue where calling a state setter in useTracker during the first run (causing an immediate re-render) will not break the hook. There was a bug there, and its fixed now.

I think that might be enough to close this. What do you think, @edemaine?

edemaine:
Wow, if you fixed the bug, that's an even better resolution to #317! Thanks!

PR #321, opened 2021-02-16
Fix data loss with immediate rerender, and add comparator argument

CaptainN:
This adds a comparator argument as the 2nd or third argument.

It also fixes an issue that can occur when an immediate rerender occurs. This can happen when a state setter is called during render. In such cases, React immediately runs the rerender without additional lifecycle stages.

This PR contains tests and readme edits for everything.

### DIFF
--- a/packages/react-meteor-data/package.json
+++ b/packages/react-meteor-data/package.json
@@ -1,11 +1,12 @@
 {
   "name": "react-meteor-data",
   "dependencies": {
-    "react": "16.13.1",
-    "react-dom": "16.13.1",
-    "react-test-renderer": "16.13.1",
     "@testing-library/react": "^10.0.2",
     "@types/meteor": "^1.4.66",
-    "@types/react": "^16.9.34"
+    "@types/react": "^16.9.34",
+    "fast-deep-equal": "^3.1.3",
+    "react": "16.13.1",
+    "react-dom": "16.13.1",
+    "react-test-renderer": "16.13.1"
   }
 }
--- a/packages/react-meteor-data/package.json
+++ b/packages/react-meteor-data/package.json
@@ -5,7 +5,7 @@
     "react-dom": "16.13.1",
     "react-test-renderer": "16.13.1",
     "@testing-library/react": "^10.0.2",
-    "@types/meteor": "^1.4.42",
+    "@types/meteor": "^1.4.66",
     "@types/react": "^16.9.34"
   }
 }
--- a/packages/react-meteor-data/useTracker.ts
+++ b/packages/react-meteor-data/useTracker.ts
@@ -68,11 +68,7 @@ const useTrackerNoDeps = <T = any>(reactiveFn: IReactiveFn<T>, skipUpdate: ISkip
     refs.computation = c;
     if (c.firstRun) {
       // Always run the reactiveFn on firstRun
-      const data = reactiveFn(c);
-      if (Meteor.isDevelopment) {
-        checkCursor(data);
-      }
-      refs.trackerData = data;
+      refs.trackerData = reactiveFn(c);
     } else if (!skipUpdate || !skipUpdate(refs.trackerData, reactiveFn(c))) {
       // For any reactive change, forceUpdate and let the next render rebuild the computation.
       forceUpdate();
@@ -139,9 +135,6 @@ const useTrackerWithDeps = <T = any>(reactiveFn: IReactiveFn<T>, deps: Dependenc
     );
     // To avoid creating side effects in render, stop the computation immediately
     Meteor.defer(() => { comp.stop() });
-    if (Meteor.isDevelopment) {
-      checkCursor(refs.data);
-    }
   }, deps);
 
   useEffect(() => {
@@ -209,7 +202,9 @@ function useTrackerDev (reactiveFn, deps = null, skipUpdate = null) {
     }
   }
 
-  return useTracker(reactiveFn, deps, skipUpdate);
+  const data = useTracker(reactiveFn, deps, skipUpdate);
+  checkCursor(data);
+  return data;
 };
 
 export default Meteor.isDevelopment
--- a/packages/react-meteor-data/useTracker.ts
+++ b/packages/react-meteor-data/useTracker.ts
@@ -104,23 +104,24 @@ const useTrackerNoDeps = <T = any>(reactiveFn: IReactiveFn<T>) => {
 }
 
 const useTrackerWithDeps = <T = any>(reactiveFn: IReactiveFn<T>, deps: DependencyList): T => {
-  let [data, setData] = useState<T>();
+  const [data, setData] = useState<T>();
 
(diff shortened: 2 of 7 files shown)


packages/react-meteor-data/package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "react-meteor-data",
   "dependencies": {
+    "@testing-library/react": "^10.0.2",
+    "@types/meteor": "^1.4.66",
+    "@types/react": "^16.9.34",
+    "fast-deep-equal": "^3.1.3",
```
dburles, line 7: Just wondering why `fast-deep-equal` is a dependency?

gunn: I'm sure just for testing. Most (all?) of those look like they could be devDependencies.

dburles: That's what I was thinking however I can't see it being referenced anywhere. They should be devDependencies but since this isn't shipped as an npm package it doesn't really matter.

packages/react-meteor-data/useTracker.ts
```diff
@@ -103,71 +108,94 @@ const useTrackerNoDeps = <T = any>(reactiveFn: IReactiveFn<T>) => {
...
+    warn("array & function", "2nd and 3rd", "deps, skipUpdate",
+      `${typeof deps} & ${typeof skipUpdate}`);
+  } else {
+    if (deps && !Array.isArray(deps) && typeof deps !== "function") {
+      warn("array or function", "2nd", "deps or skipUpdate", typeof deps);
+    }
+    if (skipUpdate && typeof skipUpdate === "function") {
+      warn("function", "3rd", "skipUpdate", typeof skipUpdate);
+    }
+  }
```
gunn, line 194: This checking will be well covered by the type signatures. Is it important to advise non-ts users?

packages/react-meteor-data/package.json
CaptainN: Correct, just for testing. Technically, these are all dev dependencies, since they should not be present for release.

packages/react-meteor-data/useTracker.ts
CaptainN: Typescript checks the values in the static environment. These check the same at runtime. If a user is writing their app primarily in JS, they'll benefit from these checks. These checks are only used in Meteor's development mode.

packages/react-meteor-data/useTracker.ts
```diff
@@ -103,71 +108,94 @@ const useTrackerNoDeps = <T = any>(reactiveFn: IReactiveFn<T>) => {
...
 
-const useTrackerWithDeps = <T = any>(reactiveFn: IReactiveFn<T>, deps: DependencyList): T => {
-  let [data, setData] = useState<T>();
-
-  const { current: refs } = useRef({ reactiveFn });
+const useTrackerWithDeps = <T = any>(reactiveFn: IReactiveFn<T>, deps: DependencyList, skipUpdate: ISkipUpdate<T> = null): T => {
+  const [data, setData] = useState<T>();
+  const { current: refs } = useRef({ reactiveFn, data });
   refs.reactiveFn = reactiveFn;
+  refs.data = data;
```
afrokick, line 115: it doesn't work, because after rerender data === undefined, which leads to refs.data === undefined.

CaptainN: I probably need to change this line:
https://github.com/meteor/react-packages/blob/e30e2905f8bcfca095f5929ef9888b061b39584a/packages/react-meteor-data/useTracker.ts#L115

to something like:

```js
if (typeof refs.data !== "undefined") {
  refs.data = data;
}
```

CaptainN: We should also write a unit test for this, seems it would be easy enough

CaptainN: I wrote a test to make sure immediate rerenders (calling a state setter or similar, during render) wouldn't reset the value. Without making this change, the test passes. I could use more help understanding how to trigger the problem you see:

```js

  Tinytest.addAsync('useTracker - immediate rerender does not result in `undefined`', async function (test, completed) {
    const container = document.createElement("DIV");
    const reactiveDict = new ReactiveDict();
    let value;
    let renders = 0;
    const Test = () => {
      renders++;
      const [num, setNum] = useState(0);
      value = useTracker(() => {
        reactiveDict.setDefault('key', 'initial');
        return reactiveDict.get('key');
      });
      if (num === 0) {
        setNum(1);
      }
      return <span>{value}</span>;
    };

    test.equal(value, 'initial', 'values should be "initial" and not undefined');
    test.equal(renders, 2, "Should have rendered twice");

    // wait for useEffect
    await waitFor(() => {}, { container, timeout: 250 });

    test.equal(value, 'initial', 'values should still be "initial" after mount');
    test.equal(renders, 3, "Should have rendered thrice");

    completed();
  });
```

CaptainN: Oh! I forgot to add the deps array, heh. Once I did that, I did indeed, confirm the bug!

The test and the hook are now fixed. Thanks!

CaptainN:
Tests have been added for an important error with immediate rerenders (calling setState before mount), and for `skipUpdate`. The errors those tests revealed have been fixed.

The only thing left to do is add some documentation around `skipUpdate`, but this is otherwise ready to ship.

packages/react-meteor-data/package.json
```diff
@@ -1,11 +1,12 @@
 {
   "name": "react-meteor-data",
   "dependencies": {
+    "@testing-library/react": "^10.0.2",
+    "@types/meteor": "^1.4.66",
+    "@types/react": "^16.9.34",
```
StorytellerCZ, line 6: Shouldn't types dependencies be in `devDependencies`?

StorytellerCZ (changes requested):

StorytellerCZ (approved):

CaptainN:
There may be another thing to check for - an immediate rerender after deps change, on an already mounted component. That will probably still yield undefined in the deps version of the hook. I need to write a test for that to confirm, and then probably write a mitigation (which will be much harder).

DavidSichau:
This also fixes our issue.

CaptainN:
I updated the readme to include info about `skipUpdate`. All tests are passing. This is ready to ship.

@filipenevola This also has some important bug fixes, but since it contains a big new feature, it should probably get a minor version bump.

Changes:
- Adds a `skipUpdate` comparator option to both `useTracker` (with and without deps) and `withTracker`.
- Fixes a bug which would sometimes cause the value to get lost (specifically, when a re-render is invoked by an immediate, in-render state change).

CaptainN:
@menelike It took me a while, but I finally added that comparator option.

afrokick (approved):

afrokick:
@filipenevola any chance to review&release it on this week?

PR #332, opened 2021-05-12
useFind and useSubscribe

CaptainN:
This PR adds `useFind` and `useSubscribe` as discussed and prototyped in the `new-hooks` branch (PR #298).

The hooks are basically done. They just need docs and tests.

TODO:
- [x] Write tests
- [x] Add documentation (readme)


CaptainN:
I added some initial tests for `useFind` and update the README!

The test covers the basic `useFind` usage. It isn't super exhaustive, but it probably covers most of what people will be doing with it.

I didn't add any tests to cover `useSubscribe` but that's such a simple wrapper over `useTracker` anyway. There is a small rerender optimization having to do with the `.ready()`/`isLoading()` method it returns that should probably be tested.

The only thing I think maybe I missed is that the `useFind` hook requires a cursor to be returned - I think there was a case where maybe we wanted the cursor to optional, in case we wanted to check for loading state within the cursor factory - but I don't know how necessary that is. It might even be beneficial to skip that check altogether, and just declaratively set up the query, and let React do the "isLoading" work. I'm not sure we need to add that functionality.

Anyway, this is probably ready for either an alpha or beta release, IMHO.

yched:
the "optional cursor" bit was because you cannot conditionally call a hook (while you can conditionally do a cursor.find() in a useTracker() callback). It's not necessarily tied to the isLoading state of a previous subscription, but could be just any condition (does some prop have a value ?)

I'd think we do need this ability to conditionally return a cursor

CaptainN:
I'll add support for null return values - but it might be an anti-pattern to do conditional queries. I'll expand on that another time.

CaptainN:
I agree it'll be challenging to support async/await in useTracker. 

Here is should be relatively easy to set it up as we need it. In `useFind` we really only use the normal fetch method once, to grab initial data. It creates a cursor and does the initial (non-reactive) data fetch in `useMemo`, and after that, it just uses the observe API on the cursor for updates.

CaptainN:
I implemented and tested `null` return value for `useFind` - this should be good to go.

StorytellerCZ:
@CaptainN ready to move this to the next stage?

CaptainN:
Yep, ready for testing! Maybe a beta release?

StorytellerCZ:
Let's resolve the conflicts and then we can do a beta release.

packages/react-mongo/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
StorytellerCZ, line 1: Looks like this is coming here from another PR.

packages/react-meteor-data/README.md
```diff
@@ -231,6 +231,59 @@ export default withTracker({
 })(Foo);
 ```
 
+#### `useFind(cursorFactory, deps)` Accellerate your lists
```
StorytellerCZ, line 234: I would switch the `useFind` and `useSubscribe` docs here, so that `useSubscribe` is covered first.

StorytellerCZ (changes requested):

packages/react-mongo/.gitignore
CaptainN: oh, hmmm

packages/react-meteor-data/README.md
CaptainN: done

packages/react-mongo/.gitignore
CaptainN: left over iteration - I removed it.

CaptainN: Npm packages are required for tests to run. Node modules should never be present during publication, but we need to keep the package.json files.

A better solution to prevent publishing node_modules is .meteorignore, but I don't know if that works in packages.

packages/react-meteor-data/README.md
```diff
@@ -231,6 +231,59 @@ export default withTracker({
...
+} else {
+  return <ul>
+    {posts.map(post => <li key={post._id}>{post.title}</li>)}
+  </ul>
+}
+```
+
+#### `useFind(cursorFactory, deps)` Accellerate your lists
+
+The `useFind` hook can substantially speed up the rendering (and rerendering) of lists coming from mongo queries (subscriptions). It does this by controlling document object references. By providing a highly tailored cursor management within the hook, using the `Cursor.observe` API, `useFind` carefully updates only the object references changed during a DDP update. This approach allows a tighter use of core React tools and philosophies to turbo charge your list renders. It is a very different approach from the more general purpose `useTracker`, and it requires a bit more set up. A notable differences is that you should NOT call `.fetch()`. `useFind` requires its factory to return a `Mongo.Cursor` object. You may also return `null`, if you want to conditionally set up the Cursor.
```
filipenevola, line 256: > You may also return `null`, if you want to conditionally set up the Cursor.

I believe an example of this case would be helpful.

filipenevola (changes requested):
Nice work @CaptainN 😄 

I just suggested a new example in the docs.

packages/react-meteor-data/README.md
filipenevola: But if the option is available but not documented properly it is even worse than documenting it properly, no?

You could say that people shouldn't use but as the option is there I believe it would be good to have an example on how to use it. Otherwise the option should be removed.

What do you think @CaptainN ?

CaptainN: I believe there's an example of it in the defunct new hooks PR from a while ago. Let me see if I can dig it up.

filipenevola:
> You can't conditionally call a subscription - but `useSubscribe` is mostly a simple wrapper around `useTracker`, and there's no reason you can't just use `useTracker` for advanced cases like this.

My point is not to block this PR, it is just to understand how common this could be.

I would say this is pretty common. But if you disagree, ok :)

CaptainN:
I think it might be less common than it seems if you are organizing your code with smaller purpose built components, the way the two hooks - `useSubscribe` and `useFind` - would imply. A recommendation might be to not even include your `useSubscribe` and `useFind` hooks in the same component. These are of course patterns that we need to set and communicate.

But like I said, it's easy enough to fall back to `useTracker` in this case anyway.

We could refactor to allow the use of a factory, instead of or in addition to the non-factory hook signature we use now, if it's too confusing the way it is.

CaptainN:
I really don't know how common it would be, but I do wonder whether it's part of Meteor's philosophy to reduce friction when coding, and if so, maybe we should add the factory option to `useSubscribe`? I don't think it'd be that difficult.

StorytellerCZ:
I have bumped the versions for next beta release @filipenevola

filipenevola (approved):

filipenevola:
Published react-meteor-data@2.4.0-beta.2.

filipenevola:
I've refactored my internal `useFind`:

```js
import { Meteor } from 'meteor/meteor';
import { getLanguage } from './languages';

const subscribeCall = (publicationName, arg) => {
  const argWithLanguage = { ...(arg || {}), language: getLanguage() };
  return Meteor.subscribe(publicationName, argWithLanguage);
};

export const useFind = ({
  publicationName,
  arg,
  find,
  skip,
  shouldSkip,
  dataReturnWhenLoading,
}) =>
  useTracker(() => {
    if (skip || (shouldSkip && shouldSkip())) {
      return {
        data: [],
        skipped: true,
      };
    }

    if (publicationName) {
      const handler = subscribeCall(publicationName, arg);

      if (!handler.ready()) {
        return {
          loading: true,
          data: dataReturnWhenLoading || undefined,
        };
      }
    }

    const result = find();

    // is findOne? so no fetch
    return { data: result.fetch ? result.fetch() : result };
  });

```

to use the new hooks and they are working perfectly, now I'm calling my hook `useData`:

```js
import { getLanguage } from './languages';
import { useFind, useSubscribe } from 'meteor/react-meteor-data';

export const useData = ({
  publicationName,
  arg,
  find,
  skip,
  shouldSkip,
  dataReturnWhenLoading,
  deps = [],
}) => {
  const argWithLanguage = { ...(arg || {}), language: getLanguage() };
  const isSkip = skip || (shouldSkip && shouldSkip());
  const isLoading = useSubscribe(
    isSkip ? null : publicationName,
    argWithLanguage
  );
  const result = useFind(find, deps);

  if (isLoading()) {
    return {
      loading: true,
      data: dataReturnWhenLoading || undefined,
    };
  }

  // is findOne? so no fetch
  return { data: result.fetch ? result.fetch() : result };
};

```

You can see it in action [here](https://www.howtocreateanapp.dev/).

CaptainN:
That's a great example of functional composition. Love it.

filipenevola:
> I'm not sure `findOne` will work though - `useFind` doesn't support `findOne`. You never have to call `fetch` on the result of `useFind`. It always returns an array of items, even if there is only one. You'd probably want to detect whether it's `findOne` or `find`, then use `useFind` to get an array, and return only the first item if it's `findOne`, or the entire array if it's `find`. (Internally, `findOne` basically works this same way - it just wraps `find` and grabs the first element.)

It's working but I believe I just have one use case for `findOne` inside a React component, I will double-check it later.

filipenevola (approved):
Looks great.

filipenevola:
v2.4.0 is published now.

@CaptainN do you want to start the blog post for this version?

filipenevola:
BTW, as always, thank you for the great work.

PR #333, opened 2021-05-12
Add react-meteor-state/use-meteor-state

CaptainN:
This adds the `react-meteor-state` package! This package contains the `useMeteorState` hook, which provides a state manager like React's build in `useState` built on Meteor's `Reload` package. The result, state persists between page refreshes during development and hot-code-push!

The only downside is that you have to name each use with a globally unique name - so this is not a direct replacement for `useState`.

TODO:
- tests
- documentation (readme)

### DIFF
--- a/packages/react-meteor-state/.gitignore
+++ b/packages/react-meteor-state/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/packages/react-meteor-state/.meteorignore
+++ b/packages/react-meteor-state/.meteorignore
@@ -1,0 +1,3 @@
+node_modules
+package.json
+package-lock.json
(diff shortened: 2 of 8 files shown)


StorytellerCZ (changes requested):
Update dependencies. Will try to test this weekend to provide more feedback.

filipenevola:
@CaptainN how can we help here?

CaptainN:
This just needs documentation and tests. (Or an executive decision to skip those)

Grubba27:
How would it be an example of the use of this?
I could probably try to help to test, but creating a good readme is an easy win.
Do you have an example of using this custom hook?

CaptainN:
It's almost the same as using `useState` except you need to name the object with a globally unique string. You would use this for something like form input - then when the page refreshes due to a meteor refresh, that data should be restored. That would work through both development refreshes and during hot-code pushes.

StorytellerCZ:
Have been using this in production for over two years now without a problem. Let's get this finished please!

Issue #344, opened 2021-12-08
how to use ssr(React Router SSR )


welkinwong:
server alert some warning
```
Warning: useFind requires an instance of Mongo.Cursor. Make sure you do NOT call .fetch() on your cursor.
```

CaptainN:
Can you post your example code? That error means that your factory method does not return a Cursor, which is required for useFind to work.

welkinwong:
> Can you post your example code? That error means that your factory method does not return a Cursor, which is required for useFind to work.


ssr package: `meteor/communitypackages:fast-render`

my code:
```
const docIsReady  = useSubscribe('metas.docInfo', { docId });
const docMeta = useFind(() => MetasCollection.find({ _id: docId }), [docId]);

console.log(docMeta);

```

the code will be running on the client and server

client console:

```
[]
[]
[object]
```

server console:

```
W20211209-19:22:41.014(8)? (STDERR) Warning: useFind requires an instance of Mongo.Cursor. Make sure you do NOT call .fetch() on your cursor.
I20211209-19:22:41.014(8)? null
```

CaptainN:
Hmm, I wonder if I need to do something different for server side. Thanks for the report! I'll look in to it.

CaptainN:
A quick audit of this shows that it should be fine. I wonder if fast render is preventing a Cursor from being returned. I'll have to look more closely in to what's going on there.

edemaine:
I found the bug here.  On the client, `find` returns a `Mongo.Cursor` (defined in [minimongo/cursor.js](https://github.com/meteor/meteor/blob/devel/packages/minimongo/cursor.js) and exported in [mongo/collection.js](https://github.com/meteor/meteor/blob/devel/packages/mongo/collection.js)).  On the server, however, `find` instead returns an instance of a private `Cursor` type defined in [mongo/mongo_driver.js](https://github.com/meteor/meteor/blob/devel/packages/mongo/mongo_driver.js)).  Sadly, the latter type is not exported, but it does have distinctive fields `_mongo` and `_cursorDescription`.

This causes two bugs in `useFind` in an SSR context:
* As reported above, `checkCursor` reports warnings on the server. Not that big a deal, because it's only in development mode.
[...]

PR #350, opened 2022-01-19
Fix useFind in SSR: check for server Cursor

edemaine:
Fixes #334, as described in [this comment](https://github.com/meteor/react-packages/issues/344#issuecomment-1012334600).

Before this PR, server-side `useFind` printed a warning and returned `[]`. After this PR, the warning is gone and the correct `cursor.find()` results are returned.


CaptainN:
Thanks so much for taking the time to produce a PR! I'll review and get back to you soon.

CaptainN:
I actually think this is kind of important to fix and ship. cc: @StorytellerCZ

StorytellerCZ (approved):

packages/react-meteor-data/useFind.ts
```diff
@@ -42,8 +42,16 @@ const useFindReducer = <T>(data: T[], action: useFindActions<T>): T[] => {
...
+// Check for valid Cursor or null.
+// On client, we should have a Mongo.Cursor (defined in
+// https://github.com/meteor/meteor/blob/devel/packages/minimongo/cursor.js and
+// https://github.com/meteor/meteor/blob/devel/packages/mongo/collection.js).
+// On server, however, we instead get a private Cursor type from
+// https://github.com/meteor/meteor/blob/devel/packages/mongo/mongo_driver.js
+// which has fields _mongo and _cursorDescription.
 const checkCursor = <T>(cursor: Mongo.Cursor<T> | undefined | null) => {
-  if (cursor !== null && cursor !== undefined && !(cursor instanceof Mongo.Cursor)) {
+  if (cursor != null && !(cursor instanceof Mongo.Cursor) &&
```
renanccastro, line 53: Shouldn't be enough checking for Meteor.isServer?

edemaine: The intent here is to print a warning if the argument is not an actual Meteor/Mongo cursor. I don't see how checking `Meteor.isServer` would suffice for that.

CaptainN: I addressed this. It's good to go.

packages/react-meteor-data/useFind.ts
```diff
@@ -121,9 +129,7 @@ const useFindServer = <T = any>(factory: () => Mongo.Cursor<T> | undefined | nul
   Tracker.nonreactive(() => {
     const cursor = factory()
     if (Meteor.isDevelopment) checkCursor(cursor)
-    return (cursor instanceof Mongo.Cursor)
-      ? cursor.fetch()
-      : null
+    return cursor ? cursor.fetch() : null
```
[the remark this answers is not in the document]
CaptainN: I addressed this. Should be good to go.

CaptainN:
This is good to go. I'm not sure why the CLA thing is failing though.

CaptainN (approved):
There were some minor issues with some of the new checks, but I changed what needed changing. Should be good to go.

packages/react-meteor-data/useFind.ts
edemaine: Not a big deal, but `undefined` and `null` are the only two values that are `== null`. Reverting is also fine though; probably clearer not to use `==`.

edemaine (commented):
I'm not a fan of how `cursor?.fetch?.() || null` transforms any falsey return value of `fetch` into `null`. This feels a bit magical; maybe it hides bugs? (in non-TypeScript code) 

`cursor?.fetch?.() ?? null` seems better; then only `undefined` gets remapped.

Actually, why not just `cursor?.fetch?.()`? This returns `undefined` if `cursor` or `cursor.fetch` are undefined/uncallable. Ah, but that's not isomorphic with the client. (But neither is the current node or the suggestion above, if `fetch` returns `undefined`.)

CaptainN:
@edemaine Great point. To be completely honest, I just wasn't sure Meteor's build pipeline supports `??` but yeah, it should totally use that. Feel free to make the change, or I can do it later. Thanks!

denihs:
Hey @CaptainN and @edemaine, as soon as this last matter is figured out, let me know, and I'll release these changes. :palms_up_together:

packages/react-meteor-data/useFind.ts
CaptainN: Ah the vagaries of Javascript's type marshalling. It's probably a bit clearer with explicit checks, but that's good to know, and now that I do know, I don't have a strong opinion either way.

denihs:
Version 2.5.1 is now up :tada:

PR #359, opened 2022-04-29
Fix useTrackerNoDeps for React 18

CaptainN:
Fixes useTrackerNoDeps for React 18

Also re-adds an attempt to retain the "side-effect"
we are creating during first render, when we can,
to both versions of the hook

The tests don't work with react 18 because the testing framework requires updating, and that breaks things. It's important to get those tests working again before we ship this, because they test important edge cases.

### DIFF
--- a/packages/react-meteor-data/CHANGELOG.md
+++ b/packages/react-meteor-data/CHANGELOG.md
@@ -1,4 +1,7 @@
 # CHANGELOG
+## v2.5.0, 2022-05-02
+* Fix useTrackerNoDeps for React 18. [PR](https://github.com/meteor/react-packages/pull/359).
+
 ## v2.4.0, 2021-12-02
 * Added `useSubscribe` and `useFind` hooks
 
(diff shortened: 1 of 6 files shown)


CaptainN:
I actually fixed the tests for the old version of react, and am satisfied that if those are working for react, and some less automated tests show this works for React 18, then we should ship this without waiting for testing rework.

This should be good to go.

StorytellerCZ (approved):

denihs:
react-meteor-data@2.5.0 published!

bratelefant:
Had an issue after updating to 2.5.0 on react 16.13.1 (usual `Invalid hook call. Hooks can only be called inside the body of a function component`); downgrading `react-meteor-data` to 2.4.0 solved it. Can't really come up with a minimal reproduction of this issue, but stack trace showed that the error originated in useTracker.ts line 69 `const data = reactiveFn(c);` in the `Tracker.nonreactive` function.

Issue #365, opened 2022-09-22
Add support for React 18


gunnartorfis:
As of React 18, `ReactDOM.render` is no longer supported.
When using the library, the following warning is displayed in the console:
<img width="608" alt="image" src="https://user-images.githubusercontent.com/5333875/191731831-d2497877-9d00-40d4-be02-bfc7c92e27ea.png">

Is [this the place](https://github.com/meteor/react-packages/blob/501f0ecfbd694d38f35afde4c96b449b69f8b60e/packages/react-template-helper/react-template-helper.js#L29) where the the render() call is made? If so, I'd be happy to open a PR with a fix. It would however probably mean a breaking change for those running on older versions of React, unless we can somehow check the version?

Grubba27:
I _think_ there will be more to it. 
to this one, I guess you can use the ``React.version``  to decide how to createRoot/render not sure if this would work 
```js 
function bootstrap(node, container) {
  if (React.version > 18) { // pseudo-validation 
    return createRoot(node, container)
  }
  return ReactDOM.render(node, container);
}
```
@CaptainN has said in some issue that I could not find what was needed to support react 18. 
We can discuss what is needed and make a checklist to support this newer version if you want.

gunnartorfis:
Gotcha.
That'd be fantastic. I'm not the most experienced when it comes to this but I'd be happy to contribute in any way I can. 

Let's wait for Capt's reply.

CaptainN:
Honestly, I've never used that render template package, and just mount the react app myself using vanilla react. That package is I think for integrating with Blaze? It should be updated, if folks are using it, but I don't know much about it.

StorytellerCZ:
I'm closing this as this appears to have been solved. Feel free to re-open if that is not the case.

PR #367, opened 2022-11-02
React 18 support

gunnartorfis:


gunnartorfis:
Solves #365

packages/react-template-helper/react-template-helper.js
```diff
@@ -5,7 +5,9 @@ checkNpmVersions({
 }, 'react-template-helper');
 
 const React = require('react');
+const { createRoot } = require('react-dom/client')
```
StorytellerCZ, line 8: Won't this throw not found error on React < 18?

gunnartorfis: Good spot, probably. 
Is there a good way to test the package locally with Meteor? Something similar to yarn link. 

Wouldn't feel comfortable merging anything without testing properly. I've only verified this is working on React 18.

StorytellerCZ: I think it should be possible to move it bellow inside the function where you check for the React version.

gunnartorfis: True! That works. I've pushed that fix.

StorytellerCZ (approved):

donflopez:
can this be merged? I'm really looking forward to use React18 in my app.

Grubba27:
Will publish a new version of react template helper

Grubba27:
Screenshot 2022-11-21 at 10 09 38
published under the 0.2.15 code

PR #374, opened 2022-11-25
Fix useFind if cursor has no value

piotrpospiech:
I fixed the behavior of `useFind` to pass a test in a `useFind.tests.js`.

```js
const docs = useFind(() => returnNull ? null : TestDocs.find(), [returnNull])
```

1. `useFind` can accept `() => null` as argument. Previously it returned `null` in this scenario, so I changed the return statement because it was returning an empty array.

2. I also added setting the `didMount` flag before checking if the cursor is `null` or `undefined`. Without this, example above will return an empty array if `returnNull` will change to `false`

### DIFF
--- a/packages/react-meteor-data/useFind.ts
+++ b/packages/react-meteor-data/useFind.ts
@@ -99,16 +99,20 @@ const useFindClient = <T = any>(factory: () => (Mongo.Cursor<T> | undefined | nu
   const didMount = useRef(false)
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


radekmie (approved):

StorytellerCZ (approved):

Grubba27:
Thx as it changes some of its behavior I will launch this change in 2.6.0

Grubba27:
ppublished

radekmie:
No, it actualy restores the one that got accidentally removed in the previous release.

PR #375, opened 2022-11-25
Update tests

piotrpospiech:
It is created as a draft because there is an issue with testing package with NPM dependencies. I created the harness app to run for now only `react-meteor-data` package tests.

Previously, `api.mainModule('tests.js');` was used to add tests. However, this error occurs when it is run this way with the new Meteor version:

image

It can be fixed by using `Npm.depends`, but it cannot be made only for tests.

If the client and server code will be stored in different files, then adding tests as I made here `api.addFiles(['useFind.ts', 'useFind.tests.js'], 'client');` will be available. However, we will need to rewrite those packages completely.

Is there any way to split client and server code because the previous approach doesn't work? Now running a test on the server will throw that error.


tests/harness/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "private": true,
+  "scripts": {
+    "test": "METEOR_PACKAGE_DIRS=../../packages  meteor test-packages react-meteor-data --driver-package test-in-browser"
+  },
+  "dependencies": {
+    "@babel/runtime": "^7.17.9",
+    "@testing-library/react": "^13.4.0",
+    "meteor-node-stubs": "^1.2.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
```
radekmie, line 12: It'd be safer to pin these.

radekmie (commented):
1. Can we remove the `react-meteor-data-harness` and `react-template-helper-harness` then?
2. @zodern could you help with the npm dependencies on the server?

piotrpospiech:
> 1. Can we remove the `react-meteor-data-harness` and `react-template-helper-harness` then?

I kept them because I cannot migrate other tests due to the npm dependencies issue. I will remove them later.

nachocodoner:
Hey, thanks for working on this PR. Right now, there's no way to run tests on React packages.

I created a new PR basing on this one and updating to latest 3.x. Closing in favor of https://github.com/meteor/react-packages/pull/422

PR #376, opened 2022-11-26
fix: named exports for useTracker and withTracker

henriquealbert:
Fix this one: https://github.com/Akryum/meteor-vite/pull/1#issuecomment-1328079910

### DIFF
--- a/packages/react-meteor-data/index.js
+++ b/packages/react-meteor-data/index.js
@@ -1,14 +1,16 @@
 /* global Meteor*/
-import React from 'react';
+import React from 'react'
 
 if (Meteor.isDevelopment) {
-  const v = React.version.split('.');
+  const v = React.version.split('.')
   if (v[0] < 16 || (v[0] == 16 && v[1] < 8)) {
-    console.warn('react-meteor-data 2.x requires React version >= 16.8.');
+    console.warn('react-meteor-data 2.x requires React version >= 16.8.')
   }
 }
 
-export { default as useTracker } from './useTracker';
-export { default as withTracker } from './withTracker.tsx';
-export { useFind } from './useFind';
-export { useSubscribe } from './useSubscribe';
+import useTracker from './useTracker'
+import withTracker from './withTracker.tsx'
+export { useFind } from './useFind'
+export { useSubscribe } from './useSubscribe'
+
+export { useTracker, withTracker }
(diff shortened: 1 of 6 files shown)


packages/react-meteor-data/index.js
```diff
@@ -8,7 +8,9 @@ if (Meteor.isDevelopment) {
...
 }
 
-export { default as useTracker } from './useTracker';
-export { default as withTracker } from './withTracker.tsx';
+import useTracker from './useTracker';
+import withTracker from './withTracker.tsx';
 export { useFind } from './useFind';
 export { useSubscribe } from './useSubscribe';
+
+export { useTracker, withTracker };
```
radekmie, line 16: 1. I'd add a comment why it's needed, as a future ESLint config may suggest to revert it.
2. Why it's not a problem for `useFind` and `useSubscribe`? Maybe changing `useTracker` and `withTracker` to use named exports would help?

henriquealbert: 1. ~~Sure! I'll add that~~ if I use named exports on both we don't need a comment here
2. As far as I noticed, Vite does not handle the current implementation `export { default as useTracker }`
the following errors occur:
`'useTracker' is not exported by meteor/react-meteor-data`
`The requested module '/@id/__x00__meteor/react-meteor-data' does not provide an export named 'useTracker'`

I'll update `useTracker` and `withTracker` to use named exports. It will look clean and standardized

Thanks for the review @radekmie ❤️

packages/react-meteor-data/withTracker.tsx
```diff
@@ -8,7 +8,7 @@ type ReactiveOptions = {
   skipUpdate?: (prev: any, next: any) => boolean;
 }
 
-export default function withTracker(options: ReactiveFn | ReactiveOptions) {
+export const withTracker = (options: ReactiveFn | ReactiveOptions) => {
```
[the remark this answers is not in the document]
Grubba27: ah got it! understood. keep it as const

Grubba27:
Will release this change under the 3.0.0 namespace because this may be a breaking change for our users(not sure) I've checked our apps, and we use it like this 
```jsx
import { useFind, useSubscribe, useTracker } from 'meteor/react-meteor-data';
```
but people maybe using ``useTracker`` and withTracker as default exports. What do you think @radekmie ?

Grubba27:
Published under the 2.6.0 namespace

radekmie:
It _is_ a breaking change, but I don't think that anyone imported it as `meteor/react-meteor-data/useTracker` anyway.

PR #383, opened 2023-02-02
Solving memory leak in useTracker

### LINKED ISSUE #351: useTracker subscription not cleaned up when deps change.
``` javascript
const useSubReady = (pagination) =>
  useTracker(() => {
    const { filters, page, perPage, sort } = pagination;
    const subPagination = Meteor.subscribe("pagination.members", filters, page, perPage, sort);

    return {
      loading: !subPagination.ready(),
    };
  }, [pagination]);
```

```javascript
  const [pagination, setPagination] = useState({
    page: 0,
    perPage: 15,
    filters: {},
    sort: { createdAt: -1 },
  });
  
 const { loading: dataLoading } = useSubReady(pagination);
 
 const onClickButton = () => {
  setPagination((prev)=> ({...prev, page:prev.page + 1}))
}
```

It's happening with old subscriptions do not cleaned up and remain when `pagination` changes.
the picture below shows the subscriptions in browser when `pagination` changes.
스크린샷 2022-01-27 오후 5 44 40

how can i solve this problem?

### LINKED ISSUE #381: react-meteor-data after update to version 2.5.0 subscription is not cleaned up
Hello all,

I encountered this issue after update to version 2.5.0 of the react-meteor-data package. It very easy to reproduce. I did so far the following in my code. I use the useTracker method to trigger a subscription as soon as any depth changes. In the mean time my depth contain limit and skip values. I realize a pagination with this. So i always renew the subscription and until version 2.4.0 it worked like a charm. Since version 2.5.0 I see in the chrome meteor plugin that the subscription is never cleaned up. I creates a new subscription with a new ID on each change of skip and limit, but never clean up the subscription. And before I always found my data with the skip value of 0. Always the part of the documents I want I found in the useTracker method when calling the find().fetch(). I enclose here my code example and as I said until version 2.40 it worked fine. according to the documentation of useTracker() its written that this will be cleaned up automatically as soon as any depth is changing or the component is unmounted. But I had no luck with the new version. I tried also version 2.6.1 (latest). Same thing happen here. Maybe someone can help me with this issue and give me a hint if there was a drastic change in the useTracker() method.

Thanks in advance for any help.

Daniel

Example Code:

`
const useOrdersSubscription = () => useTracker(() => {

        if (Meteor.isDevelopment) {
            console.log("SUBSCRIPTION ORDERS")
        }

        var getCounterReady = false;
        const subscription = Meteor.subscribe("hvzOrders", { companyAccountId: user.profile.companyAccountId }, sortQuery, limit, skip, searchValueHvzOrders, status, filterHvzOrders);
        Meteor.call("getHvzOrdersCountByCompanyAccount", { companyAccountId: user.profile.companyAccountId }, sortQuery, 0, 0, searchValueHvzOrders, status, filterHvzOrders, (err, res) => {

            getCounterReady = true;
            setOrdersCount(res);

        });
        return !subscription.ready() || getCounterReady === true;
    }, [limit, skip,  JSON.stringify(sortQuery), JSON.stringify(searchValueHvzOrders), JSON.stringify(filterHvzOrders), JSON.stringify(sortDirection)])`

`
const useOrders = () => useTracker(() => {

        if (Meteor.isDevelopment) {
            console.log("TRACKER USE ORDERS")
        }

        var query = {};

        query.companyAccountId = user.profile.companyAccountId;

        if (status) {
            query.Status = { $eq: status }
        }

        if (filterHvzOrders) {
            if (filterHvzOrders.IsDeleted) {
                if (filterHvzOrders.IsDeleted.length > 0) {

                } else {
                    query.IsDeleted = false;
                }
            } else {
                query.IsDeleted = false;
            }

            if (filterHvzOrders.IsArchived) {
                if (filterHvzOrders.IsArchived.length > 0) {

                } else {
                    query.IsArchived = false;
                }
            } else {
                query.IsArchived = false;
            }
        }
       

        if (Object.keys(filterHvzOrders).length > 0) {
            let filterQuery = [];
            Object.entries(filterHvzOrders).forEach(([key, value]) => {
                // console.log(`${key} ${value}`); // "a 5", "b 7", "c 9"
                let filterElement = key;
                let filterObj = {};
                let regexFilterValue = "";

                if (typeof value === 'object' && !moment(value, moment.ISO_8601, true).isValid()) {
                    // console.log("Value Is Object")
                    let parentKey = key;
                    if (Array.isArray(value)) {

                        //     console.log("FOUND VALUE ARRAY");
                        //     console.log(value)
                        if (value.length > 0) {
                            query[parentKey] = { $in: value };
                        }

                    } else {
                        if (value) {
                            Object.entries(value).forEach(([innerKey, innerValue]) => {

                                filterElement = parentKey + "." + innerKey;
                                if (moment(innerValue, moment.ISO_8601, true).isValid()) {
                                    //item is a date value
                                    filterElement = parentKey + "." + innerKey;

                                    if (key === "StartDate") {
                                        regexFilterValue = { $gte: innerValue }
                                    } else if (key === "EndDate") {
                                        regexFilterValue = { $lte: innerValue }
                                    } else {
                                        regexFilterValue = innerValue;
                                    }
                                } else if (typeof innerValue === 'boolean') {
                                    regexFilterValue = innerValue;

                                } else if (Array.isArray(innerValue)) {
                                    if (innerValue.length > 0) {
                                        query[filterElement] = { $in: innerValue };
                                    }
                                }
                                else {
                                    regexFilterValue = new RegExp("\\b" + innerValue, 'i');
                                }

                            });
                        }

                    }
                } else if (typeof value === 'boolean') {

                    regexFilterValue = { $eq: value };

                } else {

                    if (moment(value, moment.ISO_8601, true).isValid() || typeof value === 'boolean') {
                        //item is a date value

                        if (key === "StartDate") {
                            regexFilterValue = { $gte: value }
                        } else if (key === "EndDate") {
                            regexFilterValue = { $lte: value }
                        } else {
                            regexFilterValue = value;
                        }
                    } else {
                        regexFilterValue = new RegExp("\\b" + value, 'i');
                    }
                }

                if (regexFilterValue) {
                    if (Array.isArray(regexFilterValue)) {

                        if (value.length > 0) {
                            filterObj[filterElement] = regexFilterValue;
                            filterQuery.push(filterObj);
                        }
                    } else {
                        filterObj[filterElement] = regexFilterValue;
                        filterQuery.push(filterObj);
                    }

                }

            });
            if (filterQuery.length > 0) {
                query.$and = filterQuery;
            }
        }

        if (searchValueHvzOrders !== "") {
            let searchQuery = [];
            let split = searchValueHvzOrders.match(/\w+/g);

            if (split === null) {

            } else {
                let regex = new RegExp("\\b" + searchValueHvzOrders, 'i');

                if (HvzOrders.schema._schemaKeys.length > 0) {
                    HvzOrders.schema._schemaKeys.map((item, index) => {

                        let tmpObj = {};

                        if(item.endsWith(".$")){
                            var newItem = item.replace(".$", "");
                            tmpObj[newItem] = { $in: [regex] };
                          }else{
                            tmpObj[item] = regex;
                          }

                       // tmpObj[item] = regex;

                        searchQuery.push(tmpObj)

                    }
                    )
                }
                query.$or = searchQuery;
            }
        }

        if (Meteor.isDevelopment) {
            console.log(HvzOrders.find(query, { sort: sortQuery, limit: limit, skip: 0 }).fetch());
        }

        return HvzOrders.find(query, { sort: sortQuery, limit: limit, skip: 0 }).fetch();
    }, [limit, skip, JSON.stringify(sortQuery), JSON.stringify(searchValueHvzOrders), JSON.stringify(filterHvzOrders), JSON.stringify(sortDirection)])`

### LINKED ISSUE #382: Serious perf / mem leak in useTracker (WithDeps) >= 2.5.0
Seen with react-meteor-data >= 2.5.0 (2.4 0 is fine), react 17.2 or 18.0 :
useTracker(fn, deps) updates too frequently, causing perofrmance and memory leaks that accumulate over the component lifetime.

More precisely : 
when its reactive dependencies change, then the reactive function reruns N times in a row (each causing a component update)
with N = 1 + [number of times the useTracker deps have changed since the component was mounted]

Simple reproduction code : 

```js
import React from 'react';
import { ReactiveVar } from 'meteor/reactive-var';

const myVar = new ReactiveVar('A');

export default function ReactMeteorDataTest() {
  const [dep, setDep] = React.useState(0);
  const changeDep = () => setDep(d => d + 1);

  const value = useTracker(() => {
    console.log('re-run');
    return myVar.get();
  }, [dep]);
  const changeValue = () => myVar.set(value === 'A' ? 'B' : 'A');

  return (
    <div>
      Dependency:&nbsp;{dep}<br/>
      Reactive Value:&nbsp;{value}<br/>
      <button onClick={changeDep}>Change dependency</button>
      <button onClick={changeValue}>Change reactive value</button>&nbsp;
    </div>
  );
}
```
"Reactive Value" toggles between 'A' and 'B' on each click on "Change reactive value" button
"Dependency" is a counter that increments on each click on "Change dependency"

--> Each click on "Change reactive value" logs the 're-run' message [1 + counter] times

Grubba27:
Video for reproducing the issue as shown in #382

https://user-images.githubusercontent.com/70247653/216427576-11237efd-bcc9-468d-812e-3aa61273ecdb.mov

closes #382
closes #381
closes #351

### DIFF
--- a/packages/react-meteor-data/package.js
+++ b/packages/react-meteor-data/package.js
@@ -3,7 +3,7 @@
 Package.describe({
   name: 'react-meteor-data',
   summary: 'React hook for reactively tracking Meteor data',
-  version: '2.6.1',
+  version: '2.6.2',
   documentation: 'README.md',
   git: 'https://github.com/meteor/react-packages',
 });
--- a/packages/react-meteor-data/useTracker.ts
+++ b/packages/react-meteor-data/useTracker.ts
@@ -148,6 +148,13 @@ const useTrackerWithDeps = <T = any>(reactiveFn: IReactiveFn<T>, deps: Dependenc
         }
       })
     );
+
+    // Stop the computation immediately to avoid creating side effects in render.
+    // refers to this issues:
+    // https://github.com/meteor/react-packages/issues/382
+    // https://github.com/meteor/react-packages/issues/381
+    if (refs.comp) refs.comp.stop();
+
     // In some cases, the useEffect hook will run before Meteor.defer, such as
     // when React.lazy is used. This will allow it to be stopped earlier in
     // useEffect if needed.


CaptainN:
I actually tried this before, with no luck. I wonder if this would break the implementation on older versions of react?

Since you are stopping the computation in-line, you no longer need the Meteor.defer block within the same Memo block. That should probably be removed.

Grubba27:
> I wonder if this would break the implementation on older versions of react?

I wonder why it would. I can give it a try.

> Since you are stopping the computation in-line, you no longer need the Meteor.defer block within the same Memo block. That should probably be removed.

I thought that having both of them sounded safer (I'm not sure why I thought that), probably because it stops the computation as possible, that being inline(first if) or when setTimeout is ran

CaptainN:
You want to be careful about running that on a timeout though, because if a new one gets added to the ref before the timeout (which is always a minimum of 4ms, and I think longer in some cases with defer), you could end up cleaning up the wrong reference. (That SHOULDN'T happen, but it's not super clear and hard to test, and if you are already successfully stopping it, you might as well remove the uncertainty.)

radekmie:
I'm glad my idea worked 😅 As for what @CaptainN said, I think we could also save this computation in a constant and then stop it in `Meteor.defer`. We may revisit it if someone reports an issue.

PR #385, opened 2023-02-09
Use find async in server side using suspense

Grubba27:
as commented in the thread from this PR #360 
We should work on making server-side fetching async. In this PR, I introduce the suspense API; 

The `useFind` with suspense was intended to be used with SSR, as the client side `useFind` uses a sync function so it was not needed any changes in it


packages/react-meteor-data/react-meteor-data.d.ts
```diff
@@ -26,3 +26,7 @@ export function useSubscribe(name?: string, ...args: any[]): () => boolean;
 
 export function useFind<T>(factory: () => Mongo.Cursor<T>, deps?: React.DependencyList): T[];
 export function useFind<T>(factory: () => Mongo.Cursor<T> | undefined | null, deps?: React.DependencyList): T[] | null;
+
+export function usePromise<FN extends <T>(...args: any) => T & Promise<T>>
+(promise: (...inputs: Parameters<FN>) => Promise<ReturnType<FN>>,
+ inputs: Parameters<FN>, lifespan: number): Awaited<ReturnType<FN>>
```
radekmie, line 32: I don't think we should make it a part of our public API and rather keep it as an implementation detail. If needed, you can mark it as `@internal` in its JSDoc so TypeScript won't generate types for it.

packages/react-meteor-data/useFind.ts
```diff
@@ -51,7 +52,7 @@ const useFindReducer = <T>(data: T[], action: useFindActions<T>): T[] => {
 // which has fields _mongo and _cursorDescription.
 const checkCursor = <T>(cursor: Mongo.Cursor<T> | Partial<{ _mongo: any, _cursorDescription: any }> | undefined | null) => {
   if (cursor !== null && cursor !== undefined && !(cursor instanceof Mongo.Cursor) &&
-      !(cursor._mongo && cursor._cursorDescription)) {
+    !(cursor._mongo && cursor._cursorDescription)) {
```
radekmie, line 55: Side note: is this the official Meteor formatting? I'm not against it or something, but it'd be best not to readjust it in every PR.

(Note that the newly added import uses `"` but the rest of the code uses `'`.)

packages/react-meteor-data/useFind.ts
```diff
@@ -144,6 +145,10 @@ const useFindServer = <T = any>(factory: () => Mongo.Cursor<T> | undefined | nul
   Tracker.nonreactive(() => {
     const cursor = factory()
     if (Meteor.isDevelopment) checkCursor(cursor)
+    // in the future this maybe the default behavior
+    // for now we will use the suspense option to enable this
+    const [{suspense, suspenseLifespan}] = deps;
+    if (suspense) return usePromise(() => cursor?.fetchAsync?.(), [], suspenseLifespan || 0) ?? null
```
radekmie, line 151: 1. It should be a separate, optional argument, after the `deps`. With this implementation, `[undefined]` in `deps` will break the app.
2. What is `suspenseLifespan` for? I see that it behaves like a timeout on the promise, but I wouldn't add it now and focus on testing the `fetchAsync` alone.

packages/react-meteor-data/usePromise.ts
```diff
@@ -0,0 +1,65 @@
...
+                  promiseCaches.splice(index, 1);
+                }
+              }, lifespan);
+            }
+          }),
+      inputs,
+    };
+    promiseCaches.push(promiseCache);
+    throw promiseCache.promise;
+  };
```
radekmie, line 64: 1. In our case, `inputs` is always `[]`, so all `useFind`s will return the same result. **That's very bad.** I think we could store the cursor reference instead, but that'd require testing.
2. As argumented above, we don't need the `lifespan` argument (now).
3. As our functions have no arguments, the type could be as simple as `<T>(fn: () => Promise<T>)`. Or going even further, it could be specialized for cursors (i.e., `T` would be the document type).

radekmie (changes requested):
Let's discuss this on our next meeting.

radekmie:
I spent some time now trying things out, and I think we won't be able to implement it while preserving the API surface. The problem is that as soon as we `throw` the promise, all hooks lose their state (i.e., `.current` of `useRef` is `undefined`, `useMemo` and `useState` are initialized again, etc.). That means we'd either have to make the cursors comparable (that won't work because of `$where` and `transform`) or the API has to look differently.

Grubba27:
Hey @radekmie, I've done and found a way(hacky, I guess) to be sure to remove keys from the cached selector when it is null.
And done the same for useSubscribe with suspense as well

packages/react-meteor-data/react-meteor-data.d.ts
```diff
@@ -24,5 +26,12 @@ export function withTracker<TDataProps, TOwnProps>(options: {
 
 export function useSubscribe(name?: string, ...args: any[]): () => boolean;
 
+// Suspense
+export function useSubscribe(name: string, ...params: EJSON[]) : Meteor.SubscriptionHandle;
```
radekmie, line 30: It overloads the existing hook and thus returns a function or handle... It's incorrect, as it should be in a `/suspense` module, not in the top level (same goes for the new `useFind`).

packages/react-meteor-data/suspense/useFind.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+  }
+const useFindSuspense = <T = any>(
+  collection: Mongo.Collection<T>,
+  findArgs: Parameters<Mongo.Collection<T>['find']> | null,
+) => {
+  useEffect(() => {
+    const cachedSelectors = selectorsCache.get(collection) ?? [];
+    const _args = clone(findArgs)
```
radekmie, line 32: Do we need to `clone` it? I remember we did that during testing, but I don't think it's needed.

packages/react-meteor-data/suspense/useFind.ts
```diff
@@ -0,0 +1,126 @@
...
+          findArgs === null && cachedSelectors.length > 0
+        ) {
+          for (const selector of cachedSelectors) {
+            removeFromArray(cachedSelectors, selector);
+          }
+          selectorsCache.set(collection, cachedSelectors);
+        }
+      }, 0)
+    }
+  }, [findArgs])
```
radekmie, line 65: It removes all of the selectors whenever one of them becomes `null` - that's incorrect. For example, if I have two components and only one of them becomes `null`, the other shouldn't be affected.

packages/react-meteor-data/suspense/useFind.ts
```diff
@@ -0,0 +1,126 @@
...
+  if (typeof collection !== 'object') {
+    warn("Mongo Collection", "1st", "reactiveFn", collection);
+  }
+
+  return useFindSuspense(collection, findArgs);
+}
+
+export default Meteor.isDevelopment
+  ? useFindDev
+  : useFind;
```
radekmie, line 126: I think that we should export `useFindSuspense` on the server, but something like the existing `useFindClient` on the client - otherwise the client version won't be reactive.

radekmie (commented):
It'd be best to use some ESLint config, as the code is rather untidy.

packages/react-meteor-data/suspense/useFind.ts
Grubba27: Yeah, I could only think of using a key to solve this, but we would need three parameters...
As we talked about, cursors are not serializable.
I tried having a id being created every call of useFind but that was not helpful at all. Will try to push a useFind with this key ideia

packages/react-meteor-data/suspense/useFind.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (entry != null) ++entry.counter
+
+    removeNullCaches(cacheMap)
+    return () => {
+      // defer
+      setTimeout(() => {
+        const cachedEntries = cacheMap.get(collection)
+        const entry = cachedEntries?.find(x => isEqual(x.findArgs, findArgs))
+
+        if ((entry != null) && --entry.counter === 0) removeFromArray(cachedEntries, selector)
```
radekmie, line 48: We can make it simpler, as `removeNullCaches` will clean it anyway.
```suggestion
        if (entry != null) --entry.counter === 0;
```

packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -0,0 +1,63 @@
+import { useEffect } from 'react'
+import { EJSON } from 'meteor/ejson'
+import { Meteor } from 'meteor/meteor'
+import isEqual from 'lodash/isEqual'
+import { removeFromArray } from './useFind'
```
radekmie, line 5: This should be in a common `utils` files. Or even better, we could use `remove` from Lodash.

packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -0,0 +1,63 @@
...
+  result?: Meteor.SubscriptionHandle
+  error?: unknown
+}
+
+export function useSubscribeSuspense(name: string, ...params: EJSON[]) {
+  const cachedSubscription =
+    cachedSubscriptions.find(x => x.name === name && isEqual(x.params, params))
+
+  useEffect(() => {
+    return () => {
```
radekmie, line 23: Minor stylistic change:
```suggestion
  useEffect(() => 
    () => {
```

packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -0,0 +1,63 @@
...
+      setTimeout(() => {
+        if (cachedSubscription != null) {
+          cachedSubscription?.handle?.stop()
+          removeFromArray(cachedSubscriptions, cachedSubscription)
+        }
+      }, 0)
+    }
```
radekmie, line 30: This, similarly to `useFind`, should search for the cache entry here and not rely on the previously found one.

packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -0,0 +1,63 @@
...
+    throw cachedSubscription.promise
+  }
+
+  const subscription: SubscriptionEntry = {
+    name,
+    params,
+    promise: new Promise<Meteor.SubscriptionHandle>((resolve, reject) => {
+      const h = Meteor.subscribe(name, ...params, {
+        onReady() {
+          subscription.result = h
```
radekmie, line 45: I'm just thinking whether the handle is actually useful here - it's always going to be ready, and it doesn't make sense to `stop` it. Maybe we should not return it in the first version and see if it's needed afterwards?

packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -0,0 +1,63 @@
...
+          subscription.handle = h
+          resolve(h)
+        },
+        onStop(error: unknown) {
+          subscription.error = error
+          subscription.handle = h
+          reject(h)
```
radekmie, line 52:
```suggestion
          reject(error)
```

radekmie (changes requested):

packages/react-meteor-data/suspense/useSubscribe.ts
Grubba27: I think the same. I only need it to stop when the component unmounts. I can hide it from the end user with the `.handle` and return just null for the end user

packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -0,0 +1,64 @@
...
+    cachedSubscriptions.find(x => x.name === name && isEqual(x.params, params))
+
+  useEffect(() =>
+    () => {
+      setTimeout(() => {
+        const cachedSubscription =
+          cachedSubscriptions.find(x => x.name === name && isEqual(x.params, params))
+        if (cachedSubscription != null) {
+          cachedSubscription?.handle?.stop()
+          remove(cachedSubscriptions, x => isEqual(x, cachedSubscription))
```
radekmie, line 29: (minor) This could use `===` instead of `isEqual`, as we have a direct reference to the array's element.

radekmie (approved):
Only a couple of minor remarks. Besides that, please make sure that the PR description will stay in align with the final implementation!

packages/react-meteor-data/suspense/useFind.ts
CaptainN: What is the cache trying to do? The collection system behind Meteor.find already has caching built in.

radekmie: Let's continue this discussion in https://github.com/meteor/react-packages/pull/387#issuecomment-1470815298.

packages/react-meteor-data/suspense/useFind.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+const useFindSuspense = <T = any>(
+  collection: Mongo.Collection<T>,
+  findArgs: Parameters<Mongo.Collection<T>['find']> | null,
+  deps: React.DependencyList = []
+) => {
+  useEffect(() => {
+    const cachedEntries = cacheMap.get(collection)
+    const entry = cachedEntries?.find(x => isEqual(x.findArgs, findArgs))
+    if (entry != null) ++entry.counter
```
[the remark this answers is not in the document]
Grubba27: was eslint error but not is is solved 😄

packages/react-meteor-data/suspense/useSubscribe.ts
Grubba27: But `x` and `cachedSubscription` are both objects, that would not work as expected

Grubba27:
react-meteor-data@2.7.0 is out

packages/react-meteor-data/suspense/useSubscribe.ts
radekmie: It would, as `find` returns the references to the array element.

rj-david:
@Grubba27, can you help point to the documentation related to Suspense?

Grubba27:
I'm trying to understand....
I had not implemented the HOC for useTracker with suspense... yet.
For example, you can look with [WithTracker.tsx HOC](https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/withTracker.tsx); just remember to use `React.Suspense` and I think is possible to be done

rj-david:
Yes, I was looking at `withTracker` and it seems complicated than necessary.

Thanks @CaptainN @Grubba27

PR #387, opened 2023-03-10
Feat: useTracker with suspense

Grubba27:
## Pointed to useFind with suspense because the ESLint was already configured there.

Still missing a few pieces (it works, but it has some things that need to be solved)

I will list the issue that I think needs to be solved before this goes out:
- [x] cache deletion with a timeout of 10 (using 0 did make a recursion and things broke 😢 )
- [x] skipUpdate(refs.trackerData, data) for example now uses promises as its parameters, I'm not sure if this needs to be handled or how could I test it.

<details>
  <summary>For testing this change</summary>

  ## client

  ```ts
import { EJSON } from 'meteor/ejson';
import { Meteor } from 'meteor/meteor';
import React, { Fragment, Suspense, useCallback, useEffect, useMemo, useState } from 'react';
import { LinksCollection } from '../api/links';

import isEqual from 'lodash/isEqual';

const cachedSubscriptions: SubscriptionEntry[] = [];
type SubscriptionEntry = {
  params: EJSON[];
  name: string;
  handle?: Meteor.SubscriptionHandle;
  promise: Promise<unknown>;
  result?: Meteor.SubscriptionHandle;
  error?: unknown;
};
const removeFromArray =
  <T>(list: T[], obj: T): void => {
    if (obj) {
      const index = list.indexOf(obj);
      if (index !== -1) list.splice(index, 1);
    }
  }
function useSubscribeSuspense(name: string, ...params: EJSON[]) {
  const cachedSubscription = cachedSubscriptions.find(x => x.name === name && isEqual(x.params, params));

  useEffect(() => {
    return () => {
      setTimeout(() => {
        if (cachedSubscription) {
          cachedSubscription?.handle?.stop();
          removeFromArray(cachedSubscriptions, cachedSubscription);
        }
      }, 0)
    }
  }, [name, ...params])

  if (cachedSubscription) {
    if ('error' in cachedSubscription) {
      throw cachedSubscription.error;
    }
    if ('result' in cachedSubscription) {
      return cachedSubscription.result as Meteor.SubscriptionHandle;
    }
    throw cachedSubscription.promise;
  }

  const subscription: SubscriptionEntry = {
    name,
    params,
    promise: new Promise<void>((resolve, reject) => {
      const h = Meteor.subscribe(name, ...params, {
        onReady() {
          subscription.result = h;
          subscription.handle = h
          resolve();
        },
        onStop(error: unknown) {
          subscription.error = error;
          subscription.handle = h
          reject();
        },
      });
    }),
  };

  cachedSubscriptions.push(subscription);

  throw subscription.promise;
}




function UseTrackerClientSuspense({ active }: { active: boolean }) {
  useSubscribeSuspense('example', 10000);
  const docs = useTrackerSuspense('name',() => active ? LinksCollection.find().fetchAsync() : null ,[active]);
  const docs2 = useTracker(() => active ? LinksCollection.find().fetch() : null, [active]);

// here should be always equal
  console.log({
    docs,
    docs2
  });
  return <pre>{ JSON.stringify(docs, null, 2) }</pre>;
}


export function App() {
  const [active, setActive] = useState(false);
  const onActive = useCallback(() => setActive(active => !active), []);

  return (
    <Fragment key={ String(active) }>
      <button onClick={ async () => {
        await Meteor.callAsync('insertLink', { title: "some text", url: 'URLLLL' })
      }
      }> Add link
      </button>
      <button onClick={ onActive }>{ String(active) }</button>
      <Suspense fallback={ <div>Loader...</div> }>
        { <UseTrackerClientSuspense active={ active }/> }
      </Suspense>
    </Fragment>
  );
}



  ```

## Server

```js
import { Meteor } from 'meteor/meteor';
import { check } from 'meteor/check';
import { LinksCollection } from '/imports/api/links';

async function insertLink({ title, url }) {
  await LinksCollection.insertAsync({ title, url, createdAt: new Date() });
}

Meteor.publish('example', (limit: unknown) => {
  check(limit, Number);
  return LinksCollection.find({}, { limit });
});
Meteor.methods({
  insertLink
})
Meteor.startup(async () => {
  if (await LinksCollection.find().countAsync() === 70) LinksCollection.remove({})
  // If the Links collection is empty, add some data.
  if (await LinksCollection.find().countAsync() === 0) {
    await insertLink({
      title: 'Do the Tutorial',
      url: 'https://www.meteor.com/tutorials/react/creating-an-app',
    });

    await insertLink({
      title: 'Follow the Guide',
      url: 'https://guide.meteor.com',
    });

    await insertLink({
      title: 'Read the Docs',
      url: 'https://docs.meteor.com',
    });

    await insertLink({
      title: 'Discussions',
      url: 'https://forums.meteor.com',
    });
  }
});

```
</details>

### DIFF
--- a/packages/react-meteor-data/README.md
+++ b/packages/react-meteor-data/README.md
@@ -15,6 +15,10 @@ This package provides an integration between React and [`Tracker`](https://atmos
   - [`useSubscribe`](#usesubscribesubname-args-a-convenient-wrapper-for-subscriptions)
   - [`useFind`](#usefindcursorfactory-deps-accelerate-your-lists)
   - [Concurrent Mode, Suspense, and Error Boundaries](#concurrent-mode-suspense-and-error-boundaries)
+  - [Suspendable version of hooks](#suspendable-hooks)
+    - [useTracker](#suspense/useTracker)
+    - [useSubscribe](#suspense/useSubscribe)
+    - [useFind](#suspense/useFind)
   - [Version compatibility notes](#version-compatibility-notes)
 
 ## Install
@@ -328,13 +332,109 @@ const docs = useFind(() => {
 
 ### Concurrent Mode, Suspense and Error Boundaries
 
-There are some additional considerations to keep in mind when using Concurrent Mode, Suspense and Error Boundaries, as each of these can cause React to cancel and discard (toss) a render, including the result of the first run of your reactive function. One of the things React developers often stress is that we should not create "side-effects" directly in the render method or in functional components. There are a number of good reasons for this, including allowing the React runtime to cancel renders. Limiting the use of side-effects allows features such as concurrent mode, suspense and error boundaries to work deterministically, without leaking memory or creating rogue processes. Care should be taken to avoid side effects in your reactive function for these reasons. (Note: this caution does not apply to Meteor specific side-effects like subscriptions, since those will be automatically cleaned up when `useTracker`'s computation is disposed.)
+There are some additional considerations to keep in mind when using Concurrent Mode, Suspense and Error Boundaries, as
+each of these can cause React to cancel and discard (toss) a render, including the result of the first run of your
+reactive function. One of the things React developers often stress is that we should not create "side-effects" directly
+in the render method or in functional components. There are a number of good reasons for this, including allowing the
+React runtime to cancel renders. Limiting the use of side-effects allows features such as concurrent mode, suspense and
+error boundaries to work deterministically, without leaking memory or creating rogue processes. Care should be taken to
+avoid side effects in your reactive function for these reasons. (Note: this caution does not apply to Meteor specific
+side-effects like subscriptions, since those will be automatically cleaned up when `useTracker`'s computation is
+disposed.)
 
-Ideally, side-effects such as creating a Meteor computation would be done in `useEffect`. However, this is problematic for Meteor, which mixes an initial data query with setting up the computation to watch those data sources all in one initial run. If we wait to do that in `useEffect`, we'll end up rendering a minimum of 2 times (and using hacks for the first one) for every component which uses `useTracker` or `withTracker`, or not running at all in the initial render and still requiring a minimum of 2 renders, and complicating the API.
+Ideally, side-effects such as creating a Meteor computation would be done in `useEffect`. However, this is problematic
+for Meteor, which mixes an initial data query with setting up the computation to watch those data sources all in one
+initial run. If we wait to do that in `useEffect`, we'll end up rendering a minimum of 2 times (and using hacks for the
+first one) for every component which uses `useTracker` or `withTracker`, or not running at all in the initial render and
+still requiring a minimum of 2 renders, and complicating the API.
 
-To work around this and keep things running fast, we are creating the computation in the render method directly, and doing a number of checks later in `useEffect` to make sure we keep that computation fresh and everything up to date, while also making sure to clean things up if we detect the render has been tossed. For the most part, this should all be transparent.
+To work around this and keep things running fast, we are creating the computation in the render method directly, and
+doing a number of checks later in `useEffect` to make sure we keep that computation fresh and everything up to date,
+while also making sure to clean things up if we detect the render has been tossed. For the most part, this should all be
+transparent.
 
-The important thing to understand is that your reactive function can be initially called more than once for a single render, because sometimes the work will be tossed. Additionally, `useTracker` will not call your reactive function reactively until the render is committed (until `useEffect` runs). If you have a particularly fast changing data source, this is worth understanding. With this very short possible suspension, there are checks in place to make sure the eventual result is always up to date with the current state of the reactive function. Once the render is "committed", and the component mounted, the computation is kept running, and everything will run as expected.
+The important thing to understand is that your reactive function can be initially called more than once for a single
+render, because sometimes the work will be tossed. Additionally, `useTracker` will not call your reactive function
+reactively until the render is committed (until `useEffect` runs). If you have a particularly fast changing data source,
+this is worth understanding. With this very short possible suspension, there are checks in place to make sure the
+eventual result is always up to date with the current state of the reactive function. Once the render is "committed",
+and the component mounted, the computation is kept running, and everything will run as expected.
+
+
+## Suspendable version of hooks
+
+### `suspense/useTracker`
+
+This is a version of `useTracker` that can be used with React Suspense. It is similar to `useTracker`, 
+but it throws a promise and suspends the rendering until the promise is resolved. Also in this version it has the need
+for a key as its first parameters. the key is used to identify the computation and to avoid recreating it when the 
+component is re-rendered.
+
+```jsx
+import { useTracker } from 'meteor/react-meteor-data/suspense'
+
+function Tasks() { // this component will suspend
+  useSubscribe("tasks");
+  const { username } = useTracker("user", () => Meteor.user())  // Meteor.user() is now async
+  const tasksByUser = useTracker("tasksByUser", () =>
+          TasksCollection.find({username}, { sort: { createdAt: -1 } }).fetchAsync() // fetch will be async
+  );
+
+
+  // render the tasks
+}
+```
+
+### `suspense/useSubscribe`
+
+This is a version of `useSubscribe` that can be used with React Suspense. 
+It is similar to `useSubscribe`, it throws a promise and suspends the rendering until the promise is resolved.
+It does not return a Meteor Handle to control the subscription
+
+```jsx
+
+import { useTracker } from 'meteor/react-meteor-data/suspense'
+import { useSubscribe } from 'meteor/react-meteor-data/suspense'
+
+function Tasks() { // this component will suspend
+  useSubscribe("tasks");
+  const { username } = useTracker("user", () => Meteor.user())  // Meteor.user() is now async
+  const tasksByUser = useTracker("tasksByUser", () =>
+          TasksCollection.find({username}, { sort: { createdAt: -1 } }).fetchAsync() // fetch will be async
+  );
+
+
+// render the tasks
+}
+
+```
+
+### `suspense/useFind`
+
+This is a version of `useFind` that can be used with React Suspense.
+It has a few differences from the `useFind` without suspense, it throws a promise and suspends the rendering until the promise is resolved.
+It returns the result and it is reactive.
+You should pass as the first parameter the collection where is being searched upon and as the second parameter an array with the arguments,
+the same arguments that you would pass to the `find` method of the collection, third parameter is optional, and it is dependency array object.
+It is intended to be used in the server(SSR)
+
+```jsx
+
+import { useFind } from 'meteor/react-meteor-data/suspense'
+import { useSubscribe } from 'meteor/react-meteor-data/suspense'
+
+function Tasks() { // this component will suspend
+  useSubscribe("tasks");
+  const tasksByUser = useFind(
+          TasksCollection,
+          [{}, { sort: { createdAt: -1 } }] 
+  );
+
+  // render the tasks
+}
+
+```
+
 
 ### Version compatibility notes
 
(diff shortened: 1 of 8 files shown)


packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,305 @@
...
+export type ISkipUpdate<T> = <T>(prev: T, next: T) => boolean
+
+interface TrackerRefs {
+  computation?: Tracker.Computation
+  isMounted: boolean
+  trackerData: any
+}
+
+function resolveAsync(key: string, promise: Promise<unknown>, deps: DependencyList = []) {
+  const cached = cache.get(key) ?? false
```
radekmie, line 53: What's this `false` used for?

packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,305 @@
...
+      'Warning: your reactive function is returning a Mongo cursor. ' +
+      'This value will not be reactive. You probably want to call ' +
+      '`.fetch()` on the cursor before returning it.'
+    )
+  }
+}
+
+export const cache = new Map<string, Cache>()
+
+interface Cache {
```
radekmie, line 30: It's a _cache entry_, not _cache_. Plus it'd be nice to have a similar naming across all suspense hooks.

packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,305 @@
...
+          entry.error = error
+          reject(error)
+        })
+        .finally(() => {
+          // if 0 recursion :/
+          // also forcing an update thigs when out of hand
+          setTimeout(() => {
+            cache.delete(key)
+          }, 10)
+        })
```
radekmie, line 87: And is it needed? I don't understand what for, actually.

packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,305 @@
...
+  // Use Tracker.nonreactive in case we are inside a Tracker Computation.
+  // This can happen if someone calls `ReactDOM.render` inside a Computation.
+  // In that case, we want to opt out of the normal behavior of nested
+  // Computations, where if the outer one is invalidated or stopped,
+  // it stops the inner one.
+  Tracker.nonreactive(() =>
+    Tracker.autorun((c: Tracker.Computation) => {
+      refs.computation = c
+
+      const data: Promise<any> = Tracker.withComputation(c, async () => await reactiveFn(c))
```
radekmie, line 117: This is equivalent.
```suggestion
      const data: Promise<any> = Tracker.withComputation(c, async () => reactiveFn(c))
```

packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,305 @@
...
+      if (c.firstRun) {
+        // Always run the reactiveFn on firstRun
+        refs.trackerData = data
+        /// TODO : this about this case
+      } else if (!skipUpdate || !skipUpdate(refs.trackerData, data)) {
+        // For any reactive change, forceUpdate and let the next render rebuild the computation.
+        forceUpdate()
```
radekmie, line 124: I think this should be called only after the `promise` resolves, but I'm not sure. This'd need to be tested at least in three ways:
1. `async` computation without `await`.
2. `async` computation with `await` and a reactive source _before_ `await`.
3. `async` computation with `await` and a reactive source _after_ `await`.

packages/react-meteor-data/suspense/useTracker.ts
Grubba27: After a while, we do need to remove the old cache entry and trigger a rerender, don't we? without this piece, it is not reactive

packages/react-meteor-data/suspense/useTracker.ts
Grubba27: in 
```js
 const cached = cache.get(key) ?? false
    useEffect(() => {
    return () => {
      setTimeout(() => {
        if (cached && isEqual(cached.deps, deps)) cache.delete(key)
        //  ˆˆˆˆˆ here is false in case we do not have it; 
      }, 0)
    }
  }, [cached, key, ...deps])
```
sure it could be left as undefined, but explicitly showing that it can be false sounds better. What do you think?

packages/react-meteor-data/suspense/useTracker.ts
Grubba27: so for the tests I got the following results:

1. using the following code:

```js
  const docs = useTrackerSuspense('name', () => active ? LinksCollection.find().fetchAsync() : null, [active]);
  const docs2 = useTracker(() => active ? LinksCollection.find().fetch() : null, [active]);

  console.log({
    docs,
    docs2
  });
```
it was always reactive, every time there is a new link it would add to the docs variable.

2. using the following code:

```js
  const docs = useTrackerSuspense('name', async () => {
    const d = await LinksCollection.find().fetchAsync()
    const placeholders = await fetch('https://jsonplaceholder.typicode.com/todos').then(x => x.json())
    return active ? d : null
  }, [active]);
  const docs2 = useTracker(() => active ? LinksCollection.find().fetch() : null, [active]);

  console.log({
    docs,
    docs2
  });
```
it was always reactive, every time there is a new link it would add to the docs variable.

3. using the following code:

```js
 const docs = useTrackerSuspense('name', async () => {
    const placeholders = await fetch('https://jsonplaceholder.typicode.com/todos').then(x => x.json())
    const d = await LinksCollection.find().fetchAsync()
    return active ? d : null
  }, [active]);
  const docs2 = useTracker(() => active ? LinksCollection.find().fetch() : null, [active]);

  console.log({
    docs,
    docs2
  });

```

It was not reactive and did not change the docs variable, even when you wrap it with a Tracker.withComputation, it still needs to become reactive, and in our case, it does not .

Grubba27:
I've updated the `skipUpdate` func as well. I've opted to maintain its signature as `<T>(prev: T, next: T) => boolean` and added await to the parameters like this since we are in an async ``Tracker.autorun`` things should be fine:

```js

!skipUpdate(await refs.trackerData, await data)

```

packages/react-meteor-data/suspense/useTracker.ts
Grubba27: Actually, I was wrong in this one. The setTimout + remove cache key was supposed to be in the return of the value

packages/react-meteor-data/README.md
```diff
@@ -328,13 +332,109 @@ const docs = useFind(() => {
...
+This is a version of `useTracker` that can be used with React Suspense. It is similar to `useTracker`, 
+but it throws a promise and suspends the rendering until the promise is resolved. Also in this version it has the need
+for a key as its first parameters. the key is used to identify the computation and to avoid recreating it when the 
+component is re-rendered.
+
+```jsx
+import { useTracker } from 'meteor/react-meteor-data/suspense'
+
+function Tasks() { // this component will suspend
+  useSubscribe("tasks");
```
radekmie, line 377: I'd add an import for `useSubscribe` as well.

packages/react-meteor-data/README.md
```diff
@@ -328,13 +332,109 @@ const docs = useFind(() => {
...
+  const { username } = useTracker("user", () => Meteor.user())  // Meteor.user() is now async
```
radekmie, line 378: The "is now async" suggest _now_, but it's true only in Meteor 3.0.

packages/react-meteor-data/README.md
```diff
@@ -328,13 +332,109 @@ const docs = useFind(() => {
...
+
+
+## Suspendable version of hooks
+
+### `suspense/useTracker`
+
+This is a version of `useTracker` that can be used with React Suspense. It is similar to `useTracker`, 
+but it throws a promise and suspends the rendering until the promise is resolved. Also in this version it has the need
+for a key as its first parameters. the key is used to identify the computation and to avoid recreating it when the 
+component is re-rendered.
```
radekmie, line 371: Overall it's fine, but I just think that maybe we should focus more on _why_ and not _how_ - parts like "it throws a promise" are an implementation detail of the Suspense for data fetching (it's still experimental!). Also, please review it for grammar, e.g., "the key is used [...]" should start with a capital letter.

packages/react-meteor-data/README.md
```diff
@@ -328,13 +332,109 @@ const docs = useFind(() => {
...
+```
+
+### `suspense/useFind`
+
+This is a version of `useFind` that can be used with React Suspense.
+It has a few differences from the `useFind` without suspense, it throws a promise and suspends the rendering until the promise is resolved.
+It returns the result and it is reactive.
+You should pass as the first parameter the collection where is being searched upon and as the second parameter an array with the arguments,
+the same arguments that you would pass to the `find` method of the collection, third parameter is optional, and it is dependency array object.
+It is intended to be used in the server(SSR)
```
radekmie, line 419: That's not entirely true, I think. It's _meant_ for the SSR, but one can use it nevertheless. And the other way round - you don't have to use it if you're not interested in SSR.

packages/react-meteor-data/suspense/useTracker.ts
radekmie: I think it'd be better to leave it as `undefined` and either leave the condition as is, or add `!== undefined` there. `Entry | false` is a weird type.

packages/react-meteor-data/suspense/useTracker.ts
radekmie: 1. and 2. are OK then. As for the 3., I guess we'd need to make sure it works with `Tracker.withComputation` - I think it should. Let's discuss it tomorrow.

packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,303 @@
...
+  // Use Tracker.nonreactive in case we are inside a Tracker Computation.
+  // This can happen if someone calls `ReactDOM.render` inside a Computation.
+  // In that case, we want to opt out of the normal behavior of nested
+  // Computations, where if the outer one is invalidated or stopped,
+  // it stops the inner one.
+  Tracker.nonreactive(() =>
+    Tracker.autorun(async(c: Tracker.Computation) => {
+      refs.computation = c
+
+      const data: Promise<any> = Tracker.withComputation(c, async() => await reactiveFn(c))
```
radekmie, line 117: As before, `await` is not needed here.

packages/react-meteor-data/suspense/useTracker.ts
```diff
@@ -0,0 +1,303 @@
...
+      if (c.firstRun) {
+        // Always run the reactiveFn on firstRun
+        refs.trackerData = data
+      } else if (!skipUpdate || !skipUpdate(await refs.trackerData, await data)) {
```
radekmie, line 121: I'm not sure about these `await`s, but it's hard to come up with a direct problem too. 🤷

radekmie (approved):
Again, a couple of rather minor comments. I'm much less confident about it compared to `useFind`/`useSubscribe`, but I think it's enough for a pre-release.

CaptainN:
What is the cache system for? Meteor.find (and it's collection/pub/sub system in general) already has caching (it basically IS a client side cache). Adding a cache to useFind seems like duplicated effort.

radekmie:
The point is, we need a stable reference to a `Promise` we throw (Suspense for data fetching works like this). And as the `fetchAsync` creates a new one every single time, we need an additional "cache" here, to map the hook params into the stable reference.

CaptainN:
Ah, that's one of the constraints that prevented me from getting something like this working from inside the useFind/useTracker hook.

But I think that's not the right strategy anyway. I think the better strategy is to hold the reference where it is caught, in a provider. This is similar to how SSR (or static gen) has always worked, except now you'd be using suspense. In this case, your Provider would also be the ErrorBoundary that catches the thrown promise. (If I remember how it all goes together - been a while.)

The challenge comes from nested, derivative calls. This was a challenge to get right in Loadable too.

/2cents

packages/react-meteor-data/suspense/useTracker.ts
Grubba27: It was eslint rulemy bad :(, now is solved

PR #395, opened 2023-05-31
Gets React root from Blaze View instead of Template instance

### LINKED ISSUE #394: React components are not unmounted when Blaze destroys the parent template
As you can see in this reproduction the [clean up](https://github.com/filipenevola/blaze-react/blob/main/ui/Table.js#L16) function from useEffect is not called when Blaze is destroying the template.

filipenevola:
- fixes #394


denihs (approved):

Grubba27:
react-template-helper@0.2.17 is out  🚀

PR #414, opened 2024-11-13
Fix for #413

### LINKED ISSUE #413: Suspense throws undefined error
We upgrade from a very old projekt ([Openki](https://gitlab.com/Openki/Openki)). We started with blaze and iron router and now migrate to react, async await and meteor v3 step by step. ([Working Branch](https://gitlab.com/Openki/Openki/-/commits/dev-meteorv3))

Currently we found out that we need `react-meteor-data/suspense` to support some async await calls on loading data.

If a change happend in the UI the render breaks and [`if ('error' in cachedSubscription) throw cachedSubscription.error`](https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/suspense/useSubscribe.ts#L38) is thrown with a `undefined`. 

I gues it happens because we have mixed up a lot of diffrend librarys with diffrent Tracker in it and something is not update nice. We use blaze, react, react in blaze, iron-router (fixed for v3),

ToastHawaii:
I found out if i remove the cached subscription things works well, not sure if it is a good idea.  #413

### DIFF
--- a/packages/react-meteor-data/suspense/useSubscribe.ts
+++ b/packages/react-meteor-data/suspense/useSubscribe.ts
@@ -47,7 +47,7 @@ export function useSubscribeSuspense(name: string, ...params: EJSON[]) {
           resolve(h)
         },
         onStop(error: unknown) {
-          if(!error){
+          if (!error) {
             remove(cachedSubscriptions,
               x =>
                 x.name === subscription.name &&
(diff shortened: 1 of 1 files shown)


packages/react-meteor-data/suspense/useSubscribe.ts
```diff
@@ -51,6 +47,12 @@ export function useSubscribeSuspense(name: string, ...params: EJSON[]) {
           resolve(h)
         },
         onStop(error: unknown) {
+          if(!error){
```
Grubba27, line 50:
```suggestion
          if (!error) {
```

radekmie (commented):
Looks OK, but I'm not sure about two things:
1. Why is it limited to `!error`?
2. Can we reproduce it somehow? It'd be great to cover such an edge case with a test.

PR #416, opened 2024-12-25
Fix useSubscribe suspense does not work at SSR

welkinwong:
The useSubscribe suspense throws an error when run in the server environment, the code is similar to a normal useSubscribe and is intended to run only on the client.

```ts
const useSubscribeSuspenseServer = (name?: string, ...args: any[]) => undefined;

export const useSubscribeSuspense = Meteor.isServer
? useSubscribeSuspenseServer
: useSubscribeSuspenseClient
```

### DIFF
--- a/packages/react-meteor-data/suspense/useSubscribe.ts
+++ b/packages/react-meteor-data/suspense/useSubscribe.ts
@@ -15,7 +15,7 @@ interface Entry {
   error?: unknown
 }
 
-export function useSubscribeSuspense(name: string, ...params: EJSON[]) {
+const useSubscribeSuspenseClient = (name: string, ...params: EJSON[]) => {
   const cachedSubscription =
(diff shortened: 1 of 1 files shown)


StorytellerCZ:
This feels like a stop gap measure. Is there any functionality that can be added on the server?

welkinwong:
> This feels like a stop gap measure. Is there any functionality that can be added on the server?

server does not need Meteor.subscribe（**Client Only**）
https://docs.meteor.com/api/meteor.html#Meteor-subscribe
image

but in server-side rendering, React TSX needs to render from the server side

StorytellerCZ (approved):

StorytellerCZ:
@welkinwong you are right, I was thinking about data retrieval.

Grubba27:
looks good! I'll merge and do a new release soon.

Grubba27:
Published react-meteor-data@3.0.3.

Issue #418, opened 2025-02-25
Concurrency issue with useFind


Poyoman39:
I have found a problem with the current implementation of useFindClient.

A `fetchData` is run synchronously at first render, then useEffect run `cursor.observe({` which watch for changes on cursor.

On my component there is a concurrency issue with changes happening between the first render and the useEffect (so between the first fetchData and the observer creation)

PedroMarianoAlmeida:
I was able to create a test that catch this error:


```
// File -> packages/react-meteor-data/useFind.tests.js
Tinytest.addAsync(
    'useFind - Immediate update before effect registration (race condition test)',
    async function (test, completed) {
      const container = document.createElement('div');
      document.body.appendChild(container);

      const TestDocs = new Mongo.Collection(null);
      // Insert a single document.
      TestDocs.insert({ id: 1, val: 'initial' });

      const Test = () => {
        const docs = useFind(() => TestDocs.find(), []);
        return (
          <div data-testid="doc-value">
            {docs && docs[0] && docs[0].val}
          </div>
        );
      };

      // Render the component.
      ReactDOM.render(<Test />, container);

      // Immediately update the document (this should occur
      // after the synchronous fetch in the old code but before the effect attaches).
      TestDocs.update({ id: 1 }, { $set: { val: 'updated' } });

      // Wait until the rendered output reflects the update.
      await waitFor(() => {
        const node = container.querySelector('[data-testid="doc-value"]');
        if (!node || !node.textContent.includes('updated')) {
          throw new Error('Updated value not rendered yet');
        }
      }, { container, timeout: 500 });

      test.ok(
        container.innerHTML.includes('updated'),
        'Document should display updated value; the old code would fail to capture this update.'
      );

      document.body.removeChild(container);
      completed();
    }
  );
```

@Poyoman39 and I tried to create solutions but doesn't look like there are good ones
But at least now it is easy to reproduce

StorytellerCZ:
Probably best to add a test like this to #419.

Poyoman39:
@Grubba27 this issue should remain opened ? To me #428 only implement tests for this issue

PedroMarianoAlmeida:
I believe this Issue should stay open, and then @Poyoman39  PR https://github.com/meteor/react-packages/pull/419 should be rebased with master and add my test

(I propose a fix, but it was not good, so I kept only the test)

Poyoman39:
[...]

PR #419, opened 2025-02-25
useSyncEffect proposal Fix for #418

Poyoman39:
This intends to fix https://github.com/meteor/react-packages/issues/418
(Note that this is an early code proposal, juste to check if the idea is ok)

I think the "useSyncEffect" approach could solve the problem with useTracker leading to double rendering.


nachocodoner:
Thank you for your contribution. Can you check if the tests pass after your changes? We may automate this as a GitHub action.

Test files are in that package, specifically for useFind:
[useFind tests](https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/useFind.tests.js).

Also, consider expanding or modifying test coverage if needed for any new expectations introduced by these changes.

Poyoman39:
Thank you for your feedback @nachocodoner. I tried and failed to run tests on my machine. Is there any documentation on how to run them ?

Grubba27:
What failures did you have? 

you ran using `meteor test-packages ./path/to/package --driver-package meteortesting:mocha` ?

Poyoman39:
Thank you @Grubba27 for your help, i'm more a react dev than a meteor one for now.
I ran `meteor test-packages ./packages/react-meteor-data --driver-package meteortesting:mocha` locally, this output this error: `Cannot find module "/node_modules/meteor/react-meteor-data/node_modules/react/index.js". Try installing the npm package or make sure it is not a devDependency`

My `react-meteor-data` are successfuly installed

Poyoman39:
I still encounter this issue:

```
[[[[[ Tests ]]]]]

=> Started proxy.
=> Started MongoDB.
 (STDERR) packages/core-runtime.js:189
 (STDERR)             throw error;
 (STDERR)             ^
 (STDERR)
 (STDERR) Error: Cannot find module "/node_modules/meteor/react-meteor-data/node_modules/react/index.js". Try installing the npm package or make sure it is not a devDependency.
```

I'm using meteor 3 on my computer .. could this be the reason ?

nachocodoner:
You merged it from master, the branch with tests fixed is `tests-3.x`. To run them, once you `git checkout tests-3.x` you should go `cd tests/react-meteor-data-harness` and run `meteor npm test`.

Poyoman39:
Still running into this problem: 
debug

PedroMarianoAlmeida:
I have another approach for this issue:
https://github.com/meteor/react-packages/pull/423

radekmie (commented):
I checked the code now and I'm not sure I get the idea:
* Calling `fetchData` on mount in `useEffect`/`useSyncEffect` should not be needed, as the initial one is already covered in the initializer of `useReducer`.
* `useSyncEffect` is not really synchronous - it calls the `effect` function is `useEffect`'s cleanup _or_ in a `setTimeout`. And even if it'd be solved, I'm afraid it'd result in double render with (i.e., a flash of the old data and then the correct one).

Poyoman39:
Hi @radekmie thank you for your review :)

- You are right, I guess with my solution we would not need anymore the first fetchData provided to useReducer
- UseMemo is run synchronously at component render time, so useSyncEffect is synchronous (only cleanup is asynchronous). Since `const cleanup = useMemo(effect, deps);` manage the effect, it won't be run a second time. Based on my tests i don't have a second render.

radekmie:
Ah, right, I overlooked that it's a `useMemo`, sorry! And as for the second render, I lost track of how does triggering state change (here: `dispatch`) works in different React versions - maybe it's correct now.

Poyoman39:
No problem, thank you for your review anyway :) I hope i finally find a way to run those tests :D

packages/react-meteor-data/useFind.ts
```diff
@@ -71,6 +71,27 @@ const fetchData = <T>(cursor: Mongo.Cursor<T>) => {
...
+    }, 1000);
+  }
+
+  useEffect(() => {
+    clearTimeout(cleanupTimoutIdRef.current);
+
+    return () => {
+      cleanup();
+    };
+  }, []);
```
make-github-pseudonymous-again, line 91:
```suggestion
  }, [cleanup]);
```

packages/react-meteor-data/useFind.ts
```diff
@@ -71,6 +71,27 @@ const fetchData = <T>(cursor: Mongo.Cursor<T>) => {
...
+const useSyncEffect = (effect, deps) => {
+  const cleanup = useMemo(effect, deps);
+  const cleanupTimoutIdRef = useRef(null);
+
+  if (!cleanupTimoutIdRef.current) { // first run
+    // Clean fallback if useEffect never called
+    cleanupTimoutIdRef.current = setTimeout(() => {
+      cleanup();
+    }, 1000);
+  }
```
make-github-pseudonymous-again, line 83:
```suggestion
  const [cleanup, timeout] = useMemo(
    () => {
      const cleanup = effect();
      const timeout = setTimeout(cleanup, 1000);
      return [cleanup, timeout];
    },
    deps
  );
```

packages/react-meteor-data/useFind.ts
```diff
@@ -71,6 +71,27 @@ const fetchData = <T>(cursor: Mongo.Cursor<T>) => {
...
+
+  useEffect(() => {
+    clearTimeout(cleanupTimoutIdRef.current);
+
+    return () => {
+      cleanup();
+    };
```
make-github-pseudonymous-again, line 90:
```suggestion
    return cleanup;
```

Poyoman39:
Thank you for your review @make-github-pseudonymous-again those are nice code change suggestions :) 

Funny pseudo btw ^^

packages/react-meteor-data/useFind.ts
```diff
@@ -71,6 +71,24 @@ const fetchData = <T>(cursor: Mongo.Cursor<T>) => {
...
+      return [cleanup, timeout];
+    },
+    deps
+  );
+
+  useEffect(() => {
+    clearTimeout(cleanupTimoutIdRef.current);
+
+    return cleanup;
+  }, [cleanup]);
```
make-github-pseudonymous-again, line 88:
```suggestion
  useEffect(() => {
    clearTimeout(timeout);
    return cleanup;
  }, [cleanup, timeout]);
```

Poyoman39: Ah ah juste fixed it ^^ Thank you :)

packages/react-meteor-data/useFind.ts
```diff
@@ -71,6 +71,24 @@ const fetchData = <T>(cursor: Mongo.Cursor<T>) => {
   return data
 }
 
+const useSyncEffect = (effect, deps) => {
+  const [cleanup, timeoutId] = useMemo(
+    () => {
+      const cleanup = effect();
```
make-github-pseudonymous-again, line 77: Deduplicate cleanup calls, GC cleanup function reference, and handle `undefined` cleanup function:
```suggestion
      let _cleanup = effect();
      const cleanup = () => {
        if (_cleanup === undefined) return;
        _cleanup();
        _cleanup = undefined;
      };
```

nachocodoner:
What about adding more test coverage for this issue? [In another PR](packages/react-meteor-data/useFind.tests.js), they provided a test that seems to catch this, showing it failing without the fix and passing when applied.

StorytellerCZ:
I suggest merging #428, or at least merging that here and enabling the test.

PedroMarianoAlmeida:
The test is passing due to the early return on https://github.com/meteor/react-packages/blob/master/packages/react-meteor-data/useFind.tests.js#L113

This line should be removed on code change with the fix proposal

nachocodoner:
@PedroMarianoAlmeida: Oh, great aclaration. It is actually documented there, I misread it. Going to verify then. Thank you 🙏

nachocodoner:
Tests are passing after make them fail and merge the proposed changes here in this PR: https://github.com/meteor/react-packages/pull/419.

Going to close this one in favor of the other. As I also solved the conflicts here as well. All your work and commits are perserved.

I'm going to approve it and merge it soon for the next beta release.

nachocodoner:
This change is available in `meteor add react-meteor-data@3.0.5-beta.0`. Please, verify it works as expected.

PR #423, opened 2025-03-13
Concurrency issue with use find client

PedroMarianoAlmeida:
Possible fix for https://github.com/meteor/react-packages/issues/418

I created a new test that failed for the current `useFind` and passed for the new one

I basically removed the `useRef` and move the Initial Fetch into the Effect, this broke the **Verify reference stability between rerenders** test so I needed to update the `refresh` action on `useFindReducer` to compare the values properly (I feel that there is room from improvement on this part).

### DIFF
--- a/packages/react-meteor-data/useFind.tests.js
+++ b/packages/react-meteor-data/useFind.tests.js
@@ -107,6 +107,50 @@ if (Meteor.isClient) {
     completed()
   })
 
+  Tinytest.addAsync(
+    'useFind - Immediate update before effect registration (race condition test)',
+    async function (test, completed) {
+      const container = document.createElement('div');
+      document.body.appendChild(container);
+
+      const TestDocs = new Mongo.Collection(null);
+      // Insert a single document.
+      TestDocs.insert({ id: 1, val: 'initial' });
+
+      const Test = () => {
+        const docs = useFind(() => TestDocs.find(), []);
+        return (
+          <div data-testid="doc-value">
+            {docs && docs[0] && docs[0].val}
+          </div>
+        );
+      };
+
+      // Render the component.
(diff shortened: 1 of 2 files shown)


PedroMarianoAlmeida:
Test coverage:
Screenshot 2025-03-13 at 12 01 00 AM

My new test (Immediate update before effect registration (race condition test) with **current** `useFind`
Screenshot 2025-03-13 at 12 03 11 AM

Test code:
```
  Tinytest.addAsync(
    'useFind - Immediate update before effect registration (race condition test)',
    async function (test, completed) {
      const container = document.createElement('div');
      document.body.appendChild(container);

      const TestDocs = new Mongo.Collection(null);
      // Insert a single document.
      TestDocs.insert({ id: 1, val: 'initial' });

      const Test = () => {
        const docs = useFind(() => TestDocs.find(), []);
        return (
          <div data-testid="doc-value">
            {docs && docs[0] && docs[0].val}
          </div>
        );
      };

      // Render the component.

      // Immediately update the document (this should occur
      // after the synchronous fetch in the old code but before the effect attaches).
      TestDocs.update({ id: 1 }, { $set: { val: 'updated' } });

      // Wait until the rendered output reflects the update.
      await waitFor(() => {
        const node = container.querySelector('[data-testid="doc-value"]');
        if (!node || !node.textContent.includes('updated')) {
          throw new Error('Updated value not rendered yet');
        }
      }, { container, timeout: 500 });

      test.ok(
        container.innerHTML.includes('updated'),
        'Document should display updated value; the old code would fail to capture this update.'
      );

      document.body.removeChild(container);
      completed();
    }
  );
```

PedroMarianoAlmeida:
Sure! I will!

PR #422, opened 2025-03-12
Recover tests for latest 3.x

nachocodoner:
Continues: https://github.com/meteor/react-packages/pull/375

This PR restores the long-lost test capabilities. We didn't change much, but as reported in other PR (https://github.com/meteor/react-packages/pull/375) and the README file, tests weren’t running properly because they needed a Meteor harness app as the test runner. Existing test runners were outdated (Meteor 1.x). I updated them for Meteor 3 and confirmed they work properly with `react-meteor-data` and `react-meteor-accounts`.

With these changes, we can quickly test packages again and confidently accept new changes or features.

image

- [x] Enable GA CI for automatic test running

PR #424, opened 2025-03-13
CI support (GA runner)

nachocodoner:
Continues: https://github.com/meteor/react-packages/pull/422

This PR makes sure the CI runs the tests in this repository, providing quick feedback on changes before publishing new versions.

### DIFF
--- a/.github/workflows/test-react-packages.yml
+++ b/.github/workflows/test-react-packages.yml
@@ -21,7 +21,11 @@ jobs:
         with:
           meteor-release: '3.1.2'
 
-      - name: Cache Meteor packages
+      - name: Prepare mtest
+        run: |
+          meteor npm i -g @zodern/mtest
+
+      - name: Cache Meteor
         uses: actions/cache@v4
         with:
           path: ~/.meteor
@@ -29,14 +33,12 @@ jobs:
           restore-keys: |
             meteor-
 
-      # Install and test in Folder1 in a single step
       - name: Install and test in react-meteor-data
         working-directory: tests/react-meteor-data-harness
         run: |
           meteor npm install
           meteor npm run test:ci
 
-      # Install and test in Folder2 in a single step
       - name: Install and test in react-meteor-accounts-harness
         working-directory: tests/react-meteor-accounts-harness
         run: |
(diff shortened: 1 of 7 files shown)


.github/workflows/test-react-packages.yml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Cache Meteor
+        uses: actions/cache@v4
+        with:
+          path: ~/.meteor
+          key: meteor-cache
+
+      - name: Setup Meteor
+        uses: meteorengineer/setup-meteor@v2
+        with:
+          meteor-release: '3.1.2'
```
italojs, line 28: I suppose react-pkg runs will work with multiple Meteor versions, am I right?  

Following that, I suggest we create a Meteor version matrix to run this CI. 

``` yml
jobs:
  build:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        meteor-version: [ '3.0.0', '3.1.0', '3.1.1', '3.1.2']
      fail-fast: false
    name: Meteor ${{ matrix.meteor-version }} react-packages tests
    steps:
      - name: Checkout repository
        uses: actions/checkout@v4
      - name: Cache Meteor
        uses: actions/cache@v4
        with:
          path: ~/.meteor
          key: meteor-${{ matrix.meteor-version }}-cache
      - name: Setup Meteor
        uses: meteorengineer/setup-meteor@v2
        with:
          meteor-release: ${{ matrix.meteor-version }}

italojs (changes requested):

.github/workflows/test-react-packages.yml
nachocodoner: If we want to test this on different Meteor versions, we should also upgrade or downgrade the Meteor app used to run the tests. Upgrading is easy, downgrading less so. We can keep the test runner at the minimum Meteor version we support, making upgrades the only application on running the tests.

For now, I focused on what we actually need, and comparing tests across versions isn’t so important I think. Maybe ensuring they run in Meteor 2 is the test case, but I don’t see it as critical right now. I find it more important to ensure a matrix for React versions than for Meteor.

If this works for you, we can adjust the setup as we identify specific needs in new changes or reports.

italojs: agree

italojs (approved):

PR #428, opened 2025-03-13
Test for concurrency issue with use find client

PedroMarianoAlmeida:
Add a test that catches the issue https://github.com/meteor/react-packages/issues/418

I am skipping the test because it is failing with the current code

So this is not a fix attempt anymore, but just add the test for someone who will try to fix it

### DIFF
--- a/packages/react-meteor-data/useFind.tests.js
+++ b/packages/react-meteor-data/useFind.tests.js
@@ -106,10 +106,11 @@ if (Meteor.isClient) {
 
     completed()
   })
-  
+  // Test that catches the issue reported on https://github.com/meteor/react-packages/issues/418
   Tinytest.addAsync(
     'useFind - Immediate update before effect registration (race condition test)',
     async function (test, completed) {
+      completed(); // Remove this line to implement your change on packages/react-meteor-data/useFind.ts and check if it is working
       const container = document.createElement('div');
       document.body.appendChild(container);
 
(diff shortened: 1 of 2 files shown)


radekmie (commented):
A few comments:
1. `mergeRefreshData` does not respect order, so it'll ignore `sort` changes. That should not be ignored.
2. `mergeRefreshData` does only shallow comparisons, and that's not enough for more complex documents (nested objects are not _that_ popular, but arrays are). On the other hand, doing deep comparisons will be very expensive.
3. The way `initialData` is now calculated triggers `fetchData` on every render which is not ideal performance-wise.

PedroMarianoAlmeida:
You are right, the solution is not good after looking closely!
But at least the test was able to catch the race condition issue, so it is a starting point!

StorytellerCZ:
I'm fine with adding this test as is. But we need to find a fix for this as well.

radekmie (approved):
I agree it's a valuable test to merge.

nachocodoner:
It looks like the test here meant to highlight the issue described in https://github.com/meteor/react-packages/issues/418, but it is passing green when it should have failed. I had assumed the test would fail until https://github.com/meteor/react-packages/pull/419 was applied, after which it would pass to confirm the fix. Currently, this test is on master already, and passing properly. So supposedly is fixed without the need of any other PR to be merged.

Does anyone know what happens nad have guidance on verifying the fix using this test? I’m looking to finalize the changes for the upcoming beta release of react-packages.

PR #429, opened 2025-03-14
fix: premature cleanup in strict mode and improved performance

welkinwong:
When strict mode is enabled (development mode only), `useEffect` may run 1 to 2 times. Due to the outer layer throwing a promise, the multiple executions of `useEffect` in this scenario can lead to premature cleanup of subscriptions and cachedSubscription (i.e., before the component unmounts). Therefore, it is necessary to check the timeout to ensure that subscriptions and cachedSubscription are only cleared after the component has unmounted.

### DIFF
--- a/packages/react-meteor-data/suspense/useSubscribe.ts
+++ b/packages/react-meteor-data/suspense/useSubscribe.ts
@@ -1,10 +1,8 @@
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { EJSON } from 'meteor/ejson'
 import { Meteor } from 'meteor/meteor'
-import isEqual from 'lodash.isequal'
-import remove from 'lodash.remove'
 
-const cachedSubscriptions: Entry[] = []
+const cachedSubscriptions = new Map<string, Entry>();
 
 interface Entry {
   params: EJSON[]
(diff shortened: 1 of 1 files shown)


welkinwong:
**Supplementary Description:** Premature cleanup of subscriptions and `cachedSubscription` can lead to infinite re-mounting of components in strict mode (especially when used in conjunction with `useFind`, causing repeated execution of `useSubscribe`). This occurs because prematurely clearing `cachedSubscription` results in re-throwing a Promise, which suspends the component and creates a loop.

StorytellerCZ:
This would explain some of the issues I ran into! Looking forward to a patch release with this!

StorytellerCZ (approved):

StorytellerCZ:
Also very much appreciate the lodash removal.

welkinwong:
> I understand the problem you describe and the code makes sense. But, is there any chance to provide tests capturing the problem you describe?
> 
> I’m hesitant to merge changes in these critical packages without reliable evidence that we fully understand the issue and the fix. [Since we’ve just restored tests in this repository](https://github.com/meteor/react-packages/blob/master/tests/react-meteor-data-harness/package.json#L5), it’s a good opportunity to ensure they’re up to date and accurately cover the problem.

@nachocodoner Your argument is perfectly valid, so I've added two test cases to cover these changes. ：）

welkinwong:
This test fails on the legacy code but passes in the current PR.

Snipaste_2025-04-04_17-31-17

nachocodoner:
I didn't realize that the changes here are also included, https://github.com/meteor/react-packages/pull/430

Closing then in favor of https://github.com/meteor/react-packages/pull/435

welkinwong:
> This change is available in `meteor add react-meteor-data@3.0.5-beta.0`. Please, verify it works as expected.

Verified - normal operation 👏

PR #430, opened 2025-03-21
Fix the issue where useFind suspense does not update the cache on the server side and improved performance

welkinwong:
1. On the server, useEffect doesn't run, so cached find data isn't cleared. Even after database updates, stale data is returned on subsequent requests.
2. useFindSuspense can't run on the client (non-reactive data makes it pointless), so the useEffect part was removed.
3. lodash.remove is no longer in use and can be removed.
4. According to 7d45b72, the package-lock.json file can be deleted.

This PR builds on #429. The first commit is from the old PR, and the second commit adds new changes.


radekmie:
> useFindSuspense can't run on the client (non-reactive data makes it pointless), so the useEffect part was removed.

What do you mean by that? Of course it can! I agree, in most cases you'd go for the reactive `useFind` instead, but it's perfectly fine to use `useFindSuspense` on both client and server in isomorphic components.

welkinwong:
@radekmie Under what circumstances would one need to specifically import the non-reactive `useFindSuspense` for client-side use? The current `import { useFind } from 'meteor/react-meteor-data/suspense';` can run on both the client and server because the code includes environment detection—if it's the client, it calls the `useFindClient` code, and only if it's the server does it call `useFindSuspense`.

Or could you tell me your specific use case?

radekmie:
Ah, sorry, I misunderstood you before. In that case yes, that's a good change :+1:

welkinwong:
I've updated the PR to handle cases where the collection name is null, which was an oversight in my previous implementation.

welkinwong:
Added tests for suspense/useFind to cover the scenarios mentioned in the PR.

**before**
Snipaste_2025-04-10_12-36-15

**after**
Snipaste_2025-04-10_12-34-24

nachocodoner:
Thank you for your contribution. I am verifying the changes and tests, and as there are two PR in parallel on this work, I use the other PR I can act as a way to resolve conflicts and consolidate the changes for next beta version.

Closing in favor of https://github.com/meteor/react-packages/pull/435

PR #431, opened 2025-03-21
UseFind should work like other hooks

Poyoman39:
useFind default behavior is to have a void array as deps. This is confusing and leading to bugs hard to find.

useFind should work like useMemo, useEffect and usedTracker, having an undefined deps by default


radekmie (approved):

nachocodoner:
I didn't realize it. Ok, it may affect unexpectedly for some. I will update the changelog. Anyway, the new realese is a beta, we'll consider for the official. Good catch, thank you.

italojs:
cherrying-pick this commit to https://github.com/meteor/react-packages/pull/448

PR #438, opened 2025-04-12
fix: #443 useSyncEffect Breaks useFind Reactivity Under Suspense

welkinwong:
In the client-side implementation of `suspense/useFind`, useFindClient is used. PR #418 introduced `useSyncEffect`. 

Nevertheless, when used in **Strict Mode** and with **Suspense**, the `useSyncEffect` implementation results in non-trivial and hard-to-predict issues.

As a result, the Suspense-enabled `useFindSuspenseClient` will no longer reference `useFindClient` and will instead be a standalone implementation (derived from the last stable version of `useFindClient`).

Note: The #418 test case issue does not occur with React 18/19's `createRoot` when using `Suspense` (the test passes normally in this configuration).

### DIFF
--- a/packages/react-meteor-data/suspense/useFind.tests.js
+++ b/packages/react-meteor-data/suspense/useFind.tests.js
@@ -1,85 +1,141 @@
 /* global Meteor, Tinytest */
-import React, { Suspense } from 'react'
-import { renderToString } from 'react-dom/server'
-import { Mongo } from 'meteor/mongo'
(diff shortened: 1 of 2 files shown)


welkinwong:
Snipaste_2025-04-12_23-30-42

nachocodoner:
Merging this since I confirm the fix. Will publish a beta soon. Thanks a lot for the detailed PRs and focus on these issues.

PR #440, opened 2025-04-14
Add test to detect useless double rendering

Poyoman39:


StorytellerCZ:
I'm guessing that the test should currently fail.

PR #441, opened 2025-04-17
perf(suspense): optimize handling of multiple suspense/useSubscribe

welkinwong:
Strictly speaking, this is indeed a bug. While it doesn’t break the app, it can lead to a poor user experience in specific scenarios:

> When two or more identical subscriptions are active simultaneously, unmounting one of them may cause the other subscription’s nearest Suspense boundary to revert to its fallback state.

This issue typically occurs in complex applications.

---

This PR primarily introduces a reference counter.  The subscription will only be cleared when the counter reaches zero, indicating that no components are actively consuming it.

### DIFF
--- a/packages/react-meteor-data/suspense/useSubscribe.tests.js
+++ b/packages/react-meteor-data/suspense/useSubscribe.tests.js
@@ -29,7 +29,7 @@ if (Meteor.isServer) {
         );
       };
 
-      const { queryByText, findByText } = render(<TestSuspense />, {
+      const { queryByText, findByText, unmount } = render(<TestSuspense />, {
         container: document.createElement('container'),
       });
 
@@ -39,6 +39,8 @@ if (Meteor.isServer) {
       );
 
       test.isTrue(await findByText('nameA'), 'Need to return data');
+
+      unmount();
     }
   );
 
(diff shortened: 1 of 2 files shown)


welkinwong:
**before:**
Snipaste_2025-04-17_20-56-50

**after:**
Snipaste_2025-04-17_20-55-13

nachocodoner:
Not sure if it's related, but it might be. In the forums, [users reported a failure in react-meteor-data@3.0.5-beta.0](
https://forums.meteor.com/t/communitypackages-react-router-ssr-v6-beta-release/63418/7), including the recent changes we merged around Suspense, useFind, and useSubscribe. One user said the [react-router-ssr-demo](https://github.com/copleykj/react-router-ssr-demo) doesn’t load data at all in that beta. The demo code uses Suspense and the hooks, so it could be closely tied to those changes. Not sure if it’s covered by the fix in this PR or if another fix is needed.

Could you check? I’ll test it when I get a break from the latest bundler work.

welkinwong:
@nachocodoner 

After investigation, I believe this is an implementation issue in the v6 version of the `react-router-ssr` module.  The current implementation uses React 18's `renderToPipeableStream` with `onAllReady` for completion signaling, but the stream doesn't wait for all asynchronous data to resolve.

I've already filed an issue regarding this behavior in the react-router-ssr repository: [renderToPipeableStream in server.js Does Not Wait for Async Content Inside Suspense](https://github.com/Meteor-Community-Packages/react-router-ssr/issues/40)

nachocodoner (approved):

PR #448, opened 2025-06-11
UseFind should work like other hooks

italojs:
cherrying pick [this](https://github.com/meteor/react-packages/pull/431) PR's commit to the right branch (R4.0.0) after removing from master


nachocodoner:
I think should be included since it is used in that file code state.

nachocodoner (approved):

PR #482, opened 2026-04-28
fix: useTracker should only render once on mount (closes #439)

### LINKED ISSUE #439: useTracker renders two times at mount, would be nice if it renders only once
This code renders two times

```
    const Test = ({ afterMountInc = false }) => {
      renderCount++;
      console.log(renderCount);
      value = useTracker(() => {
        return null;
      }, []);
      return <span>{renderCount}</span>;
    };

```

nicolv:
## Summary

Closes #439 - `useTracker` renders two times at mount

## Problem

The previous implementation of `useTrackerNoDeps` would:
1. Stop and recreate the Tracker computation on **every render**
2. Combined with React 18 StrictMode double-rendering, this caused the component to render more times than expected

## Solution

1. **Added `useSyncEffect` hook** - A custom hook that runs effects synchronously during render but defers cleanup by 1 second to avoid StrictMode double-rendering issues

2. **Simplified `useTrackerNoDeps`** - Now delegates to `useTrackerWithDeps` instead of having its own complex render-phase logic

3. **Updated `useTrackerWithDeps`** - Uses `useSyncEffect` to manage the Tracker computation lifecycle, avoiding the stop/recreate on every render

## Changes

- `packages/react-meteor-data/useSyncEffect.ts` - New file: the `useSyncEffect` hook
- `packages/react-meteor-data/useTracker.ts` - Simplified to use `useSyncEffect`

## Key Technical Points

- The previous code used `Meteor.defer()` and render-phase computation creation, which caused issues with StrictMode
- `useSyncEffect` runs the effect in `useMemo` (during render) but delays cleanup by 1 second, avoiding StrictMode double-render issues
- This is the same approach as PR #436 but refined

## Testing

The existing `noDepsTester` test in `useTracker.tests.js` already covers StrictMode behavior.

---

Related to #439

### DIFF
--- a/packages/react-meteor-data/useSyncEffect.ts
+++ b/packages/react-meteor-data/useSyncEffect.ts
@@ -1,0 +1,25 @@
+import { useMemo, useEffect, DependencyList } from 'react';
+
+/**
+ * Runs effect synchronously during render (like useMemo), but defers cleanup
+ * by 1s to avoid issues with React StrictMode double-rendering.
+ * This is used for Tracker computations that should not be stopped/restarted
+ * on each render.
+ */
+const useSyncEffect = (effect: () => any, deps: DependencyList) => {
+  const [cleanup, timeoutId] = useMemo(
+    () => {
+      const cleanup = effect();
+      const timeoutId = setTimeout(cleanup, 1000);
+      return [cleanup, timeoutId];
+    },
+    deps
+  );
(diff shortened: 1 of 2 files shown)


StorytellerCZ:
Looks elegant using new hooks. I'll try to find some time to test this out.

PR #496, opened 2026-08-19
react-meteor-data: support TypeScript 7

italojs:
## Summary

Adds `typescript@7.0.2` to the explicit supported constraint used by `react-meteor-data`, and advances the pending beta version to `4.1.0-beta.1`.

## Why

Meteor’s TypeScript 7 work resolves `typescript@7.0.2`. Without this constraint, apps that use `react-meteor-data` cannot select TypeScript 7 and the dependency solver fails.

This is stacked on #490, which adds TypeScript 6 support. It also supports the Meteor TypeScript 7 work in meteor/meteor#14319.

## Validation

- Reproduced the dependency-solver conflict with TypeScript 7 before the change.
- Confirmed a temporary Meteor app can select `typescript@7.0.2` with the updated local package.
- Ran `meteor npm run test:ci` in `tests/react-meteor-data-harness`.

No package has been published as part of this PR.

### DIFF
--- a/packages/react-meteor-data/CHANGELOG.md
+++ b/packages/react-meteor-data/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## v4.1.0-beta.1, 2026-08-19
+* Add compatibility with the `typescript@7.x` Meteor package so the package can
+  be used on Meteor releases built against TypeScript 7 (see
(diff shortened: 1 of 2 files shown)


italojs:
Superseded by #497, which contains the TypeScript 7-only change directly on master.

